PR #42, opened 2014-08-17
Fixed API dependencies

core23:


core23:
As the REST bundle is not required for the this bundle, the api form and controllers definitions should only be loaded, when the bundles are used.

DependencyInjection/SonataClassificationExtension.php
```diff
@@ -47,8 +47,11 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('orm.xml');
         $loader->load('form.xml');
         $loader->load('serializer.xml');
-        $loader->load('api_controllers.xml');
-        $loader->load('api_form.xml');
+        
+        if (isset($bundles['FOSRestBundle']) && isset($bundles['NelmioApiDocBundle'])) {
```
Soullivaneuh: Not sure that `NelmioApiDocBundle` is required to get REST working.

core23:
@Soullivaneuh Any better idea? It is the same logic like in the news bundle...

core23:
ping @Soullivaneuh

Soullivaneuh:
I would like to hear @rande opinion for this.

core23:
ping @rande

Soullivaneuh:
@OskarStark please avoid to randomly assign contributors just if they answer on a PR.

Assign only if issues / PR is closely related to a specific contributor. And if it's not sure, comment an explanation of why.

BTW in this case, I would assign @rande for last review.

PR #81, opened 2015-01-01
Better context support for tags and collections

afurculita:


core23:
Fixed in https://github.com/sonata-project/SonataClassificationBundle/pull/154

PR #90, opened 2015-02-09
Minor bug fixes

srascar:
This PR contains multiple bug fixes:

 - Remove MediaInterface in setters of Category / Collection models

    This will prevent the following Doctrine exception when using ClassificationBundle without         MediaBundle
    
    *request.CRITICAL: Uncaught PHP Exception Doctrine\Common\Proxy\Exception\UnexpectedValueException: "The type hint of parameter "media" in method "setMedia" in class "Application\ClassificationBundle\Entity\Category" is invalid."*
    
 - Add context field in Category form
 
    This will prevent the following Doctrine exception when trying to insert new Category

    *request.CRITICAL: Uncaught PHP Exception RuntimeException: "Context cannot be null"*
    
 - Add required => false for enable field in category
 
    Allow to create / update disabled categories

 - Add required => false for parent field in category
 
    Allow to create root categories

### DIFF
--- a/Model/Category.php
+++ b/Model/Category.php
@@ -270,8 +270,14 @@ abstract class Category implements CategoryInterface
     /**
      * {@inheritdoc}
      */
-    public function setMedia(MediaInterface $media = null)
+    public function setMedia($media = null)
     {
+        // No hinting is used to prevent Doctrine parsing exception
+        // when SonataMediaBundle is not installed
+        if ($media !== null && !($media instanceof MediaInterface)) {
+            throw new \InvalidArgumentException('Method setMedia expects null or instance of Sonata\MediaBundle\Model\MediaInterface');
+        }
+
         $this->media = $media;
     }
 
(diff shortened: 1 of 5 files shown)


Model/Category.php
```diff
@@ -270,7 +270,7 @@ public function getParent()
     /**
      * {@inheritdoc}
      */
-    public function setMedia(MediaInterface $media = null)
+    public function setMedia($media = null)
```
Bladrak: If you remove type hinting, you should add a check in the setter for the type and explain why you did so in the comments

rande: You still need to raise an exception if the class is invalid

srascar:
done @rande

Model/Category.php
Soullivaneuh: And it's also BC break for children classes.

Soullivaneuh:
@srascar Plase see my review: https://github.com/sonata-project/SonataClassificationBundle/pull/90#discussion_r32616099

core23:
Please rebase and check comments @srascar 

Could you prevent a BC break?

Soullivaneuh:

Regards.

Issue #96, opened 2015-03-20
getRootCategories of categoryManager is not compartible with SonataMedia bundle


bardius:
Sonata media bundle context field holds a string (that is the name of the context). When the getRootCategories is called the returned categories array has as index the id of the classification context that is mapped to the categories instead, so no matching categories are found for the context of Media Bundle.

To fix this i had to amend line 130 of Entity/Categorymanager to:
$categories[$category->getContext()->getName()] = $loadChildren ? $this->getRootCategory($category->getContext()) : $category;

Replaced the index from $category->getContext()->getId() to $category->getContext()->getName()

rande:
Can you create a PR ? this will be easier to review

Thanks

bardius:
Okay, when i get time I will amend the getRootCategories function to get a second parameter for if it is used by media bundle or not and return the results array mapped by context name in this case

PR #98, opened 2015-03-20
Allow the root categories to be mapped by context name, rather than id, ...

bardius:
...to accomondate compatibility with sonata media event subscriber.

This come as a hot fix for https://github.com/sonata-project/SonataClassificationBundle/issues/96


bardius:
Apparently the context table id field is a varchar so the same value of the name can be set as an id. Thus this fix is not required

PR #101, opened 2015-03-26
Update CategoryManager.php

josephzhao:
I am using postgresql server 9.3 

$categories = $this->getObjectManager()->createQuery(sprintf('SELECT c FROM %s c WHERE c.context = :context ORDER BY c.parent ASC', $class))		
     ->setParameter('context', $context->getId())
             ->execute();

I always get the result with parent is null as last record, so always show the error

throw new \RuntimeException('The first category must be the root');

I temporally resolve my problem with code 

Thanks

Joseph

### DIFF
--- a/Entity/CategoryManager.php
+++ b/Entity/CategoryManager.php
@@ -189,10 +189,14 @@ class CategoryManager extends BaseEntityManager implements CategoryManagerInterf
 
         $class = $this->getClass();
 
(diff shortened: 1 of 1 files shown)


core23:
Would this fix your issue https://github.com/sonata-project/SonataClassificationBundle/pull/108 @josephzhao ?

Soullivaneuh:

Regards.

PR #107, opened 2015-03-31
Don't hide the field ID when create a new context

### LINKED ISSUE #106: Unable to create new Context from Admin

jerome-fix:
fixes #106


kamilZ:
+1 works like a charm

ppetermann:
this will always show the id field, which wasn't intended in the original code (and i *guess*) could cause problems when editing (as the id shouldn't be changed ;)

core23:
Please rebase @jerome-fix

OskarStark:
we cant merge this you are not allowed to change the ID later :-1:

PR #112, opened 2015-04-07
Update Context.orm.xml.skeleton

faheem00:
The id type was varchar before, which does not support auto increment. ID should be set to int, so that context can be inserted.


core23:
@faheem00 Can you explain why you want a int value for the context id?

OskarStark:
this is wrong @faheem00 the id here is a string

rande:
context is a string. this is a design choice.

PR #113, opened 2015-04-10
Added missing id getters

core23:


core23:
Like in the other bundles, the id getter should be defined in the interfaces.

This is a BC.

Soullivaneuh:
I'll let @rande review it.

There is no way to make it BC?

If not, could you please make a upgrade not for next major release?

rande:
ok to test

PR #115, opened 2015-05-02
Missing required dependency

Axinet:
Documentation mention registering JsonType, but it isn't available (results in class not found exception).


rande:
@Axinet I think it is an issue with the documention, and not with composer.json.

Axinet:
Oops. You are right. I blindly followed documentation, but in this bundle there is no requirement for json type in doctrine.

PR #135, opened 2015-09-18
Form category position

qsomazzi:
I don't understand why position should always be set to 0,

When you try to edit a category already well positionned, we don't want to move it. (or reinit it)


OskarStark:
:+1:

PR #141, opened 2015-09-29
Fix hungarian translations

sagikazarmark:


Soullivaneuh:
For my curiosity: What is the difference? Why your translations should be better?

sagikazarmark:
Actually I translated it originally. Back then I translated it to its exact meaning, but I found a word which better fits there. The word 'Kontextus' can rather be used when you are talking about some context that you can see, hear. Actually the word 'Környezet' can be translated to 'environment' in english. They have similar meaning in that language as well. But in hungarian that form fits better here I think.

Soullivaneuh:
Thanks for the knowledge moment. :+1:

PR #145, opened 2015-10-05
Line endings and CS fixes

sagikazarmark:


OskarStark:
Thank you @sagikazarmark

PR #165, opened 2015-12-02
Update CategorySelectorType.php

trihtm:
Depend on Symfony Reference. When you use version 2.7+, this options must be added, that make this bundle still work.

http://symfony.com/doc/current/reference/forms/types/choice.html#choices-as-values

```
Catchable Fatal Error: Argument 1 passed to Symfony\Component\Form\Extension\Core\Type\ChoiceType::normalizeLegacyChoices() must be of the type array, null given
```

### DIFF
--- a/Form/Type/CategorySelectorType.php
+++ b/Form/Type/CategorySelectorType.php
@@ -52,6 +52,7 @@ class CategorySelectorType extends AbstractType
             'choice_list'       => function (Options $opts, $previousValue) use ($that) {
                 return new SimpleChoiceList($that->getChoices($opts));
             },
+            'choices_as_values' => true,
         ));
     }
 


OskarStark:
hmm, but then you can use this only with symfony version >2.7 :-1: 

can you handle this a bit better?

ping @Soullivaneuh

OskarStark:
and please make a better commit message

trihtm:
The reference said: "Before 2.7, this option didn't actually exist, but the behavior was equivalent to setting this to false in 2.7."
So I think symfony version < 2.7 may work well because there was not this option.

OskarStark:
:+1:

rande:
If the option is unknown this will generates an exception for previous version. This option only exists for 2.7 and 2.8. So this PR will break SF3.0 support.

PR #183, opened 2016-04-21
Composer fixes and improves

Soullivaneuh:
**THIS BUNDLE WILL HAVE A NEW MAJOR RELEASE**

* 3.x-dev
* Use new composer constraints notation
* Remove minimum-stability option to clearly show which required packages are not stable

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "symfony/framework-bundle": "^2.3 || ^3.0",
         "symfony/form": "^2.3 || ^3.0",
-        "sonata-project/doctrine-orm-admin-bundle": "2.4.x-dev@dev",
+        "sonata-project/doctrine-orm-admin-bundle": "3.x-dev@dev",
         "sonata-project/admin-bundle": "3.x-dev@dev",
         "sonata-project/datagrid-bundle": "^2.2"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,32 +17,30 @@
...
-        "sonata-project/media-bundle": "~2.2",
-        "symfony/phpunit-bridge": "~2.7|~3.0"
+        "friendsofsymfony/rest-bundle": "^1.1",
+        "jms/serializer-bundle": "~0.11 || ^1.0",
+        "sonata-project/media-bundle": "^2.2",
+        "symfony/phpunit-bridge": "^2.7 || ^3.0"
     },
     "suggest": {
-        "sonata-project/media-bundle": "~2.2"
+        "sonata-project/media-bundle": "^2.2"
```
core23: The suggest property should contain text, no version number https://getcomposer.org/doc/04-schema.md#suggest

Soullivaneuh: I know, but this is not the goal of the PR and not the priority right now. :-)

Feel free to open a PR and ping me for suggestion. :+1:

Soullivaneuh: Note: Have versions on `conflicts` key would be great.

PR #186, opened 2016-05-04
Document, Tag and Collection manager classes don't respect PageableManagerInterface

Jean-Beru:
In order to implements [PageableManagerInterface](https://github.com/sonata-project/SonataCoreBundle/blob/2.x/Model/PageableManagerInterface.php#L31) (~2.2), `getPager` method of TagManager, CollectionManager and CategoryManager have to be updated.

### DIFF
--- a/Document/CategoryManager.php
+++ b/Document/CategoryManager.php
@@ -22,7 +22,7 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
     /**
      * {@inheritdoc}
      */
-    public function getPager(array $criteria, $page, $maxPerPage = 10)
+    public function getPager(array $criteria, $page, $limit = 10, array $sort = array())
     {
         $parameters = array();
 
@@ -37,7 +37,7 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
         $query->setParameters($parameters);
 
         $pager = new Pager();
-        $pager->setMaxPerPage($maxPerPage);
+        $pager->setMaxPerPage($limit);
         $pager->setQuery(new ProxyQuery($query));
         $pager->setPage($page);
         $pager->init();
--- a/Document/CollectionManager.php
+++ b/Document/CollectionManager.php
@@ -22,7 +22,7 @@ class CollectionManager extends BaseDocumentManager implements CollectionManager
     /**
      * {@inheritdoc}
      */
-    public function getPager(array $criteria, $page, $maxPerPage = 10)
+    public function getPager(array $criteria, $page, $limit = 10, array $sort = array())
     {
         $parameters = array();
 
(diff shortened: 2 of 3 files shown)


Document/CategoryManager.php
```diff
@@ -22,7 +22,7 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
     /**
      * {@inheritdoc}
      */
-    public function getPager(array $criteria, $page, $maxPerPage = 10)
+    public function getPager(array $criteria, $page, $limit = 10, array $sort = array())
```
Soullivaneuh: Changing method signature is BC break.

Soullivaneuh:
Please update your commit to have a proper message and description.

See: https://github.com/sonata-project/dev-kit/blob/master/doc/CONTRIBUTING.md#pull-requests

Document/CategoryManager.php
Jean-Beru: I agree but `getPager` method can't be used with the old signature.

greg0ire:
"defition" => "definition"

Document/CategoryManager.php
Soullivaneuh: @greg0ire I don't know if it should be a major or a patch for this case because this class already inherit from core-bundle class that have this signature since the begining...

ping @rande.

greg0ire: If it has never worked (how could it?) it is a patch.

Soullivaneuh: @Jean-Beru Could you tell us more about that? Since when do you have this bug?

Soullivaneuh:
@Jean-Beru I see you updated the title, but the commit message has to be changed too. :+1:

Document/CategoryManager.php
Jean-Beru: @Soullivaneuh I have this bug when Symfony cache is warming. I don't know when it appears because I worked on m'y actual project since few days.

Jean-Beru: @Soullivaneuh To be more concise, versions are :
```
classification-bundle: 2.2.x-dev (~2.2@dev in composer.json)
code-bundle: 2.3.9 (not specified in composer.json)
```

This patch have been already merged in master (see https://github.com/sonata-project/SonataClassificationBundle/commit/da9ab9c463ee4a4979d0946f47980421421136a3 and https://github.com/sonata-project/SonataClassificationBundle/commit/890ac2114618ee5824f2ef0bc7644988ac22dab6)

Jean-Beru:
@Soullivaneuh, @greg0ire As explained in https://github.com/sonata-project/SonataClassificationBundle/pull/186#discussion_r62459738, my changes are also present in master too. Maybe I should do a other PR with these "cherry-picked" commits ?

Soullivaneuh:
Sorry for the auto-close. I'll release 3.0 soon based on master. You can reopen the PR against 3.x branch when it will be available :+1:

PR #191, opened 2016-05-19
Missing the proper schema update

AndrewReborn:
As the last step of the installation process, an update to the schema is required. This quite important step is missing at the end of the doc and, in my opinion, should be added. A non expert user could not understand what's happening when get errors via the Sonata Admin Interface: in fact the actions are regularly displayed in the dashboard and the errors are triggered only the create process with the form submit

### DIFF
--- a/Resources/doc/reference/installation.rst
+++ b/Resources/doc/reference/installation.rst
@@ -101,3 +101,9 @@ Installation
                     mappings:
                         ApplicationSonataClassificationBundle: ~
                         # ...
+
+* Update the schema:
+
+.. code-block:: bash
+
+    php bin/console doctrine:schema:update --force


Resources/doc/reference/installation.rst
```diff
@@ -101,3 +101,9 @@ Installation
                     mappings:
                         ApplicationSonataClassificationBundle: ~
                         # ...
+
+* Update the schema:
+
+.. code-block:: bash
+
+    php bin/console doctrine:schema:update --force
```
Soullivaneuh: :-1: for that.

`doctrine:schema:update` is not a command to use on production for example, you should use migrations.

Soullivaneuh:

Regards.

PR #182, opened 2016-04-16
Added abstract blocks

core23:
### Changelog

```markdown
### Added
- Added `AbstractCategoriesBlockService` class
- Added `AbstractCollectionsBlockService` class
- Added `AbstractTagsBlockService` class
```

### Subject

This adds some basic blocks that can be used in other bundles.

It does nearly the same as #143, but adds some more blocks for collections and categories and has a better abstraction layer for other bundles.

### To do

- [X] Added abstract blocks for categories, collections and tags
- [X] Update the tests
- [X] Update the documentation

PR #143, opened 2015-09-30
Tag block

sagikazarmark:
Implemented a tags block. It can be configured to use different contexts (although currently it can only accept a string). Number of displayed tags can also be configured, but limitation is disabled by default.

PR #201, opened 2016-05-26
Added abstract blocks

core23:
### Changelog

```markdown
### Added
- Added `AbstractCategoriesBlockService` class
- Added `AbstractCollectionsBlockService` class
- Added `AbstractTagsBlockService` class
```

### Subject

This adds some basic blocks that can be used in other bundles.

It does nearly the same as #143, but adds some more blocks for collections and categories and has a better abstraction layer for other bundles.

Refs https://github.com/sonata-project/SonataClassificationBundle/pull/182

### To do

- [X] Added abstract blocks for categories, collections and tags
- [X] Update the tests
- [X] Update the documentation

### DIFF
--- a/Block/Service/AbstractCategoriesBlockService.php
+++ b/Block/Service/AbstractCategoriesBlockService.php
@@ -1,0 +1,189 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Block\Service;
+
+use Sonata\AdminBundle\Admin\AdminInterface;
+use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\BlockBundle\Block\BlockContextInterface;
+use Sonata\BlockBundle\Model\BlockInterface;
+use Sonata\ClassificationBundle\Model\CategoryInterface;
+use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+use Sonata\CoreBundle\Model\Metadata;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Christian Gripp <mail@core23.de>
+ */
+abstract class AbstractCategoriesBlockService extends AbstractClassificationBlockService
+{
+    /**
+     * @var CategoryManagerInterface
+     */
+    private $categoryManager;
+
+    /**
+     * @var AdminInterface
+     */
+    private $categoryAdmin;
+
+    /**
+     * @param string                   $name
+     * @param EngineInterface          $templating
+     * @param ContextManagerInterface  $contextManager
+     * @param CategoryManagerInterface $categoryManager
+     * @param AdminInterface           $categoryAdmin
+     */
+    public function __construct($name, EngineInterface $templating, ContextManagerInterface $contextManager, CategoryManagerInterface $categoryManager, AdminInterface $categoryAdmin)
+    {
+        parent::__construct($name, $templating, $contextManager);
+
+        $this->categoryManager = $categoryManager;
+        $this->categoryAdmin = $categoryAdmin;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function execute(BlockContextInterface $blockContext, Response $response = null)
+    {
+        $category = $this->getCategory($blockContext->getSetting('categoryId'), $blockContext->getSetting('category'));
+        $root = $this->categoryManager->getRootCategory($blockContext->getSetting('context'));
+
+        return $this->renderResponse($blockContext->getTemplate(), array(
+            'context' => $blockContext,
+            'settings' => $blockContext->getSettings(),
+            'block' => $blockContext->getBlock(),
+            'category' => $category,
+            'root' => $root,
+        ), $response);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildEditForm(FormMapper $formMapper, BlockInterface $block)
+    {
+        $adminField = $this->getFormAdminType($formMapper, $this->categoryAdmin, 'categoryId', 'category', array(
+            'label' => 'form.label_category',
+        ), array(
+            'translation_domain' => 'SonataClassificationBundle',
+            'link_parameters' => array(
+                array(
+                    array(
+                        'context' => $block->getSetting('context'),
+                    ),
+                ),
+            ),
+        ));
+
+        $formMapper->add('settings', 'sonata_type_immutable_array', array(
+            'keys' => array(
+                array('title', 'text', array(
+                    'label' => 'form.label_title',
+                    'required' => false,
+                )),
+                array('context', 'choice', array(
+                    'label' => 'form.label_context',
+                    'required' => false,
+                    'choices' => $this->getContextChoices(),
+                )),
+                array($adminField, null, array()),
+            ),
+            'translation_domain' => 'SonataClassificationBundle',
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureSettings(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(array(
+            'title' => 'Categories',
+            'category' => false,
+            'categoryId' => null,
+            'context' => 'default',
+            'template' => 'SonataClassificationBundle:Block:base_block_categories.html.twig',
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function load(BlockInterface $block)
+    {
+        if (is_numeric($block->getSetting('categoryId'))) {
+            $block->setSetting('categoryId', $this->getCategory($block->getSetting('categoryId')));
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function prePersist(BlockInterface $block)
+    {
+        $this->resolveIds($block);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function preUpdate(BlockInterface $block)
+    {
+        $this->resolveIds($block);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBlockMetadata($code = null)
+    {
+        $description = (!is_null($code) ? $code : $this->getName());
+
+        return new Metadata($this->getName(), $description, false, 'SonataClassificationBundle', array(
+            'class' => 'fa fa-folder-open-o',
+        ));
+    }
+
+    /**
+     * @param CategoryInterface|int  $id
+     * @param CategoryInterface|null $default
+     *
+     * @return CategoryInterface
+     */
+    final protected function getCategory($id, CategoryInterface $default = null)
+    {
+        if ($id instanceof CategoryInterface) {
+            return $id;
+        }
+
+        if (is_numeric($id)) {
+            return $this->categoryManager->find($id);
+        }
+
+        return $default;
+    }
+
+    /**
+     * @param BlockInterface $block
+     */
+    private function resolveIds(BlockInterface $block)
+    {
+        $block->setSetting(
+            'categoryId',
+            is_object($block->getSetting('categoryId')) ? $block->getSetting('categoryId')->getId() : null
+        );
+    }
+}
--- a/Block/Service/AbstractClassificationBlockService.php
+++ b/Block/Service/AbstractClassificationBlockService.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Block\Service;
+
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Sonata\AdminBundle\Admin\AdminInterface;
+use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\BlockBundle\Block\BaseBlockService;
+use Sonata\ClassificationBundle\Model\ContextInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\Form\FormBuilder;
+
+/**
+ * @author Christian Gripp <mail@core23.de>
+ */
+abstract class AbstractClassificationBlockService extends BaseBlockService
+{
+    /**
+     * @var ContextManagerInterface
+     */
+    protected $contextManager;
+
+    /**
+     * @param string                  $name
+     * @param EngineInterface         $templating
+     * @param ContextManagerInterface $contextManager
+     */
+    public function __construct($name, EngineInterface $templating, ContextManagerInterface $contextManager)
+    {
+        parent::__construct($name, $templating);
+
+        $this->contextManager = $contextManager;
+    }
+
+    /**
+     * @param FormMapper     $formMapper
+     * @param AdminInterface $admin
+     * @param string         $formField
+     * @param string         $field
+     * @param array          $fieldOptions
+     * @param array          $adminOptions
+     *
+     * @return FormBuilder
+     */
+    final protected function getFormAdminType(FormMapper $formMapper, AdminInterface $admin, $formField, $field, $fieldOptions = array(), $adminOptions = array())
+    {
+        $adminOptions = array_merge(array(
+            'edit' => 'list',
+            'translation_domain' => 'SonataClassificationBundle',
+        ), $adminOptions);
+
+        $fieldDescription = $admin->getModelManager()->getNewFieldDescriptionInstance($admin->getClass(), $field, $adminOptions);
+        $fieldDescription->setAssociationAdmin($admin);
+        $fieldDescription->setAdmin($formMapper->getAdmin());
+        $fieldDescription->setAssociationMapping(array(
+            'fieldName' => $field,
+            'type' => ClassMetadataInfo::MANY_TO_ONE,
+        ));
+
+        $fieldOptions = array_merge(array(
+            'sonata_field_description' => $fieldDescription,
+            'class' => $admin->getClass(),
+            'model_manager' => $admin->getModelManager(),
+            'required' => false,
+        ), $fieldOptions);
+
+        return $formMapper->create($formField, 'sonata_type_model_list', $fieldOptions);
+    }
+
+    /**
+     * Returns an context choice array.
+     *
+     * @return string[]
+     */
+    final protected function getContextChoices()
+    {
+        $contextChoices = array();
+        /* @var ContextInterface $context */
+        foreach ($this->contextManager->findAll() as $context) {
+            $contextChoices[$context->getId()] = $context->getName();
+        }
+
+        return $contextChoices;
+    }
+}
--- a/Block/Service/AbstractCollectionsBlockService.php
+++ b/Block/Service/AbstractCollectionsBlockService.php
@@ -1,0 +1,190 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Block\Service;
+
+use Sonata\AdminBundle\Admin\AdminInterface;
+use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\BlockBundle\Block\BlockContextInterface;
+use Sonata\BlockBundle\Model\BlockInterface;
+use Sonata\ClassificationBundle\Model\CollectionInterface;
+use Sonata\ClassificationBundle\Model\CollectionManagerInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+use Sonata\CoreBundle\Model\Metadata;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Christian Gripp <mail@core23.de>
+ */
+abstract class AbstractCollectionsBlockService extends AbstractClassificationBlockService
+{
+    /**
+     * @var CollectionManagerInterface
+     */
+    private $collectionManager;
+
+    /**
+     * @var AdminInterface
+     */
+    private $collectionAdmin;
+
+    /**
+     * @param string                     $name
+     * @param EngineInterface            $templating
+     * @param ContextManagerInterface    $contextManager
+     * @param CollectionManagerInterface $collectionManager
+     * @param AdminInterface             $collectionAdmin
+     */
+    public function __construct($name, EngineInterface $templating, ContextManagerInterface $contextManager, CollectionManagerInterface $collectionManager, AdminInterface $collectionAdmin)
+    {
+        parent::__construct($name, $templating, $contextManager);
+
+        $this->collectionManager = $collectionManager;
+        $this->collectionAdmin = $collectionAdmin;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function execute(BlockContextInterface $blockContext, Response $response = null)
+    {
+        $collection = $this->getCollection($blockContext->getSetting('collectionId'), $blockContext->getSetting('collection'));
+        $collections = $this->contextManager->findBy(array(
+            'enabled' => true,
+            'context' => $blockContext->getSetting('context'),
+        ));
+
+        return $this->renderResponse($blockContext->getTemplate(), array(
+            'context' => $blockContext,
+            'settings' => $blockContext->getSettings(),
+            'block' => $blockContext->getBlock(),
+            'collection' => $collection,
+            'collections' => $collections,
+        ), $response);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildEditForm(FormMapper $formMapper, BlockInterface $block)
+    {
+        $adminField = $this->getFormAdminType($formMapper, $this->collectionAdmin, 'collectionId', 'collection', array(
+            'label' => 'form.label_collection',
+        ), array(
+            'translation_domain' => 'SonataClassificationBundle',
+            'link_parameters' => array(
+                array(
+                    'context' => $block->getSetting('context'),
+                ),
+            ),
+        ));
+
+        $formMapper->add('settings', 'sonata_type_immutable_array', array(
+            'keys' => array(
+                array('title', 'text', array(
+                    'label' => 'form.label_title',
+                    'required' => false,
+                )),
+                array('context', 'choice', array(
+                    'label' => 'form.label_context',
+                    'required' => false,
+                    'choices' => $this->getContextChoices(),
+                )),
+                array($adminField, null, array()),
+            ),
+            'translation_domain' => 'SonataClassificationBundle',
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureSettings(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(array(
+            'title' => 'Collections',
+            'collection' => false,
+            'collectionId' => null,
+            'context' => null,
+            'template' => 'SonataClassificationBundle:Block:base_block_collections.html.twig',
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function load(BlockInterface $block)
+    {
+        if (is_numeric($block->getSetting('collectionId'))) {
+            $block->setSetting('collectionId', $this->getCollection($block->getSetting('collectionId')));
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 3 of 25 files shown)


Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -0,0 +1,195 @@
...
+    private $categoryAdmin;
+
+    /**
+     * @param string                   $name
+     * @param EngineInterface          $templating
+     * @param ContextManagerInterface  $contextManager
+     * @param CategoryManagerInterface $categoryManager
+     * @param CategoryAdmin            $categoryAdmin
+     */
+    public function __construct(
```
Soullivaneuh: One lined signature.

Block/Service/AbstractCollectionsBlockService.php
```diff
@@ -0,0 +1,196 @@
...
+    private $collectionAdmin;
+
+    /**
+     * @param string                     $name
+     * @param EngineInterface            $templating
+     * @param ContextManagerInterface    $contextManager
+     * @param CollectionManagerInterface $collectionManager
+     * @param CollectionAdmin            $collectionAdmin
+     */
+    public function __construct(
```
Soullivaneuh: One-lined signature.

Block/Service/AbstractTagsBlockService.php
```diff
@@ -0,0 +1,203 @@
...
+    private $tagAdmin;
+
+    /**
+     * @param string                  $name
+     * @param EngineInterface         $templating
+     * @param ContextManagerInterface $contextManager
+     * @param TagManagerInterface     $tagManager
+     * @param TagAdmin                $tagAdmin
+     */
+    public function __construct(
```
Soullivaneuh: One-lined signature.

Block/Service/BaseClassificationBlockService.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param FormMapper     $formMapper
+     * @param AdminInterface $admin
+     * @param string         $formField
+     * @param string         $field
+     * @param array          $fieldOptions
+     * @param array          $adminOptions
+     *
+     * @return FormBuilder
+     */
+    protected function getFormAdminType(
```
Soullivaneuh: One-lined signature.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [3.x]
+### Added
+
+### Added
+- Added `AbstractCategoriesBlockService` class
```
Soullivaneuh: Please, keep the changelog for the PR body only.

Block/Service/AbstractCategoriesBlockService.php
core23: Sure? @greg0ire said that I have to split it...

Soullivaneuh:
Some of your methods should be final, should not be?

Block/Service/AbstractCategoriesBlockService.php
greg0ire: Well in case you didn't notice, I lost [2 battles](https://github.com/sonata-project/dev-kit/issues/103) about that against @Soullivaneuh , javiereguiluz and fabpot yesterday… so just use one-line signatures, I don't want to get started on this again :disappointed: Sorry for the inconvenience.

core23:
@Soullivaneuh Could we modify the coveralls settings? IMO they are too strict. 

No idea how to improve the coverage. There are tests for this.

greg0ire:
@core23 : looks like it does not prevent us from clicking the merge button though…

core23:
But in the [PR overview](https://github.com/sonata-project/SonataClassificationBundle/pulls), it looks like this PR is incomplete or has failing tests. Don't know if anybody will have a look at *failing* PRs.

greg0ire:
Indeed, that's not ideal…

Soullivaneuh:
@core23 A coverage should not decrease at all except for a very good reason.

This is why the failing coverage should not appear a lot on PR.

Or maybe it's a false positive but I would be surprised because I used this tool from a long time and I never see this happen.

> @core23 : looks like it does not prevent us from clicking the merge button though…

@greg0ire Not for *now*. :stuck_out_tongue:

Soullivaneuh:
I just took a look of your first added class:

image

Well, this is maybe related to the decrease, isn't it?

Tests/Block/Service/AbstractCategoriesBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+            ->with($this->equalTo('categoryId'), $this->equalTo($category));
+
+        $blockService = $this->getMockForAbstractClass('Sonata\ClassificationBundle\Block\Service\AbstractCategoriesBlockService', array(
+            'block.service', $this->templating, $this->contextManager, $this->categoryManager, $this->categoryAdmin,
+        ));
+        $blockService->load($block);
+    }
+
+    /**
+     * @covers Sonata\BlockBundle\Tests\Block\AbstractCategoriesBlockService::preUpdate
```
Soullivaneuh: Please do not use `@covers`. We follow the Symfony way.

core23: What is the symfony way? Do you have any docs how to write tests that covers multiple methods? `preUpdate` and `prePersist` do the same.

Soullivaneuh: The Symfony way is to not use `@covers` at all and let the coverage cover the called code.

This was discussed on another PR. I have to update the code on dev-kit about that.

Resources/views/Block/base_block_categories.html.twig
```diff
@@ -0,0 +1,87 @@
...
+
+This file is part of the Sonata package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+
+#}
+{% extends sonata_block.templates.block_base %}
```
Soullivaneuh: Please add a new line.

Resources/views/Block/base_block_collections.html.twig
```diff
@@ -0,0 +1,37 @@
...
+
+This file is part of the Sonata package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+
+#}
+{% extends sonata_block.templates.block_base %}
```
Soullivaneuh: Please add a new line.

Resources/views/Block/base_block_tags.html.twig
```diff
@@ -0,0 +1,33 @@
...
+
+This file is part of the Sonata package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+
+#}
+{% extends sonata_block.templates.block_base %}
```
Soullivaneuh: Please add a new line.

Tests/Block/Service/AbstractCategoriesBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+namespace Sonata\ClassificationBundle\Tests\Block\Service;
+
+use Sonata\BlockBundle\Tests\Block\AbstractBlockServiceTest;
+use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
+use Sonata\ClassificationBundle\Admin\CategoryAdmin;
+use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+
+/**
+ * Class AbstractCategoriesBlockServiceTest.
```
Soullivaneuh: Useless class comment.

Tests/Block/Service/AbstractCollectionsBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+namespace Sonata\ClassificationBundle\Tests\Block\Service;
+
+use Sonata\BlockBundle\Tests\Block\AbstractBlockServiceTest;
+use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
+use Sonata\ClassificationBundle\Admin\CollectionAdmin;
+use Sonata\ClassificationBundle\Model\CollectionManagerInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+
+/**
+ * Class AbstractCollectionsBlockServiceTest.
```
Soullivaneuh: Useless comment.

Tests/Block/Service/AbstractCollectionsBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+            ->with($this->equalTo('collectionId'), $this->equalTo($collection));
+
+        $blockService = $this->getMockForAbstractClass('Sonata\ClassificationBundle\Block\Service\AbstractCollectionsBlockService', array(
+            'block.service', $this->templating, $this->contextManager, $this->collectionManager, $this->collectionAdmin,
+        ));
+        $blockService->load($block);
+    }
+
+    /**
+     * @covers Sonata\BlockBundle\Tests\Block\AbstractCollectionsBlockService::preUpdate
```
Soullivaneuh: Please remove `@covers`.

Tests/Block/Service/AbstractTagsBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+namespace Sonata\ClassificationBundle\Tests\Block\Service;
+
+use Sonata\AdminBundle\Tests\Fixtures\Admin\TagAdmin;
+use Sonata\BlockBundle\Tests\Block\AbstractBlockServiceTest;
+use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
+use Sonata\ClassificationBundle\Model\TagManagerInterface;
+
+/**
+ * Class AbstractTagsBlockServiceTest.
```
Soullivaneuh: Useless comment.

Tests/Block/Service/AbstractTagsBlockServiceTest.php
```diff
@@ -0,0 +1,121 @@
...
+            ->with($this->equalTo('tagId'), $this->equalTo($tag));
+
+        $blockService = $this->getMockForAbstractClass('Sonata\ClassificationBundle\Block\Service\AbstractTagsBlockService', array(
+            'block.service', $this->templating, $this->contextManager, $this->tagManager, $this->tagAdmin,
+        ));
+        $blockService->load($block);
+    }
+
+    /**
+     * @covers Sonata\BlockBundle\Tests\Block\AbstractTagsBlockService::preUpdate
```
Soullivaneuh: Please remove `@covers`.

core23:
Everything green 🎉 @Soullivaneuh

Soullivaneuh:
@core23 The code looks good!

But I would like a final review of someone using this bundle.

cc @sonata-project/contributors

core23:
Can you have a look at this @sagikazarmark

sagikazarmark:
Will do and will add translations here.

Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -0,0 +1,188 @@
...
+    {
+        $description = (!is_null($code) ? $code : $this->getName());
+
+        return new Metadata($this->getName(), $description, false, 'SonataClassificationBundle', array(
+            'class' => 'fa fa-folder-open-o',
+        ));
+    }
+
+    /**
+     * @param mixed $id
```
sagikazarmark: Based on the code this can be integer and some kind of object? Can we narrow it down? Is there a base category or interface? (I think there is) It would make easier to read what this method does.

Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -0,0 +1,188 @@
...
+            return $this->categoryManager->find($id);
+        }
+
+        return $default;
+    }
+
+    /**
+     * @param BlockInterface $block
+     */
+    private function resolveIds(BlockInterface $block)
```
sagikazarmark: Should this be final as well? Can still be overridden in children.

Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -0,0 +1,188 @@
...
+    private $categoryAdmin;
+
+    /**
+     * @param string                   $name
+     * @param EngineInterface          $templating
+     * @param ContextManagerInterface  $contextManager
+     * @param CategoryManagerInterface $categoryManager
+     * @param AdminInterface           $categoryAdmin
+     */
+    public function __construct($name, EngineInterface $templating, ContextManagerInterface $contextManager, CategoryManagerInterface $categoryManager, AdminInterface $categoryAdmin)
```
sagikazarmark: Shouldn't this be on multiple lines according to PSR-2?

Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -0,0 +1,188 @@
...
+
+        return $default;
+    }
+
+    /**
+     * @param BlockInterface $block
+     */
+    private function resolveIds(BlockInterface $block)
+    {
+        $block->setSetting('categoryId',
```
sagikazarmark: IMO this should be on newline too

Block/Service/AbstractClassificationBlockService.php
```diff
@@ -0,0 +1,95 @@
...
+     * @param FormMapper     $formMapper
+     * @param AdminInterface $admin
+     * @param string         $formField
+     * @param string         $field
+     * @param array          $fieldOptions
+     * @param array          $adminOptions
+     *
+     * @return FormBuilder
+     */
+    final protected function getFormAdminType(FormMapper $formMapper, AdminInterface $admin, $formField, $field, $fieldOptions = array(), $adminOptions = array())
```
sagikazarmark: Shouldn't this be on multiple lines according to PSR-2?

Block/Service/AbstractCollectionsBlockService.php
```diff
@@ -0,0 +1,189 @@
...
+            'collections' => $collections,
+        ), $response);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildEditForm(FormMapper $formMapper, BlockInterface $block)
+    {
+        $adminField = $this->getFormAdminType($formMapper, $this->collectionAdmin, 'collectionId', 'collection', array(
```
sagikazarmark: Shouldn't this be on multiple lines according to PSR-2?

Block/Service/AbstractCategoriesBlockService.php
OskarStark: `final` on a `private` method?

Block/Service/AbstractCollectionsBlockService.php
```diff
@@ -0,0 +1,189 @@
...
+    {
+        $description = (!is_null($code) ? $code : $this->getName());
+
+        return new Metadata($this->getName(), $description, false, 'SonataClassificationBundle', array(
+            'class' => 'fa fa-folder-open-o',
+        ));
+    }
+
+    /**
+     * @param mixed $id
```
sagikazarmark: Same here as above

Block/Service/AbstractCollectionsBlockService.php
```diff
@@ -0,0 +1,189 @@
...
+            return $this->collectionManager->find($id);
+        }
+
+        return $default;
+    }
+
+    /**
+     * @param BlockInterface $block
+     */
+    private function resolveIds(BlockInterface $block)
```
sagikazarmark: Same here as above

Block/Service/AbstractTagsBlockService.php
```diff
@@ -0,0 +1,196 @@
...
+    private $tagAdmin;
+
+    /**
+     * @param string                  $name
+     * @param EngineInterface         $templating
+     * @param ContextManagerInterface $contextManager
+     * @param TagManagerInterface     $tagManager
+     * @param AdminInterface          $tagAdmin
+     */
+    public function __construct($name, EngineInterface $templating, ContextManagerInterface $contextManager, TagManagerInterface $tagManager, AdminInterface $tagAdmin)
```
sagikazarmark: Shouldn't this be on multiple lines according to PSR-2?

Block/Service/AbstractCategoriesBlockService.php
greg0ire: Sadly, no… T___T I wish.

greg0ire: Related battle I lost about this are here : 

- https://github.com/symfony/symfony/issues/18890
- https://github.com/sonata-project/dev-kit/issues/103

Some support would be welcome if you want to chime in against this kind of nonsense…

greg0ire: Current arguments for oneline are : 

- it's a good reminder that the code is shitty and the number of arguments should be decreased
- there is the phpdoc if you want to see the arguments
- such a change will create merge nightmares
- we should be consistent with every decision symfony makes
- we (and by we I actually mean they) kind of like it better like this

sagikazarmark: Okay, so this optional, not mandatory then. My vote goes to being consistent too.

sagikazarmark:
Hungarian translations:

``` xml
            <trans-unit id="form.label_category">
                <source>form.label_category</source>
                <target>Kategória</target>
            </trans-unit>
            <trans-unit id="form.label_collection">
                <source>form.label_collection</source>
                <target>Gyűjtemény</target>
            </trans-unit>
            <trans-unit id="form.label_tag">
                <source>form.label_tag</source>
                <target>Címke</target>
            </trans-unit>
            <trans-unit id="form.label_title">
                <source>form.label_title</source>
                <target>Cím</target>
            </trans-unit>
            <trans-unit id="no_collections_found">
                <source>no_collections_found</source>
                <target>Gyűjtemény nem található</target>
            </trans-unit>
            <trans-unit id="no_categories_found">
                <source>no_categories_found</source>
                <target>Kategória nem található</target>
            </trans-unit>
            <trans-unit id="no_tags_found">
                <source>no_tags_found</source>
                <target>Címke nem található</target>
            </trans-unit>
```

sagikazarmark:
:+1: from me, nice work @core23

Block/Service/AbstractTagsBlockService.php
Soullivaneuh: No. http://symfony.com/doc/current/contributing/code/standards.html#structure

> Declare all the arguments on the same line as the method/function name, no matter how many arguments there are;

Plus, there is no hard limit on PSR-2, only soft.

Block/Service/AbstractCollectionsBlockService.php
Soullivaneuh: https://github.com/sonata-project/SonataClassificationBundle/pull/201#discussion_r66786510

Block/Service/AbstractClassificationBlockService.php
Soullivaneuh: https://github.com/sonata-project/SonataClassificationBundle/pull/201#discussion_r66786510

Block/Service/AbstractTagsBlockService.php
sagikazarmark: Already discussed here:

https://github.com/sonata-project/SonataClassificationBundle/pull/201#discussion-diff-66782389

Block/Service/AbstractCategoriesBlockService.php
Soullivaneuh: > Some support would be welcome if you want to chime in against this kind of nonsense…

This discussion was already made and closed, no need for more...

Block/Service/AbstractCategoriesBlockService.php
sagikazarmark: https://3v4l.org/578Rs
https://3v4l.org/VL1Js

Block/Service/AbstractCategoriesBlockService.php
greg0ire: > This discussion was already made and closed, no need for more...

So people who are late are wrong ? The only argument that makes sense in the Sonata discussion IMO is "consistency". And it does not make sense in the Symfony discussion, because if you change long method signature to multi line, then you are consistent.

On a side note, I see more and more bad things done in Sonata for the sake of consistency, no later than this morning : "well this is how it is done in SonataXXXBundle", and in this case, if you have short signatures onelined and long signatures multilined, you can't really call that "inconsistent", can you? You can only if you have some long signatures multilined and some other long signatures onelined.

Block/Service/AbstractCategoriesBlockService.php
sagikazarmark: > final on a private method? 

Why not?

Block/Service/AbstractCategoriesBlockService.php
sagikazarmark: I don't think we should discuss it here (even if I tend to agree with @greg0ire ). Also, communication between to parties can hardly be called a "discussion". But if the decision is to follow symfony instead of following logic (whatever logic it might be), then I don't think we should start convincing anyone to do otherwise.

OskarStark:
Let's merge this after @core23 applied the translations 👍

Block/Service/AbstractCategoriesBlockService.php
greg0ire: > I don't think we should start convincing anyone to do otherwise.

Apart from Symfony, but I already tried, so… lost battle.

Block/Service/AbstractCategoriesBlockService.php
Soullivaneuh: private method can't be overridden AFAIK, so no need to make them final. Am I wrong?

sagikazarmark: See the two tests above

OskarStark: oO didn't know that!

Soullivaneuh: > https://3v4l.org/578Rs

The HELL?! O_O

sagikazarmark: This is perfectly logical if you see it from a visibility point of view: the child class no nothing about the parent's private scope, so it can override anything that's private, but can't access the parent.

greg0ire: > private method can't be overridden AFAIK, so no need to make them final. Am I wrong?

lol of course it can, why not ? People can choose the naming they want in child classes, can't they ?

greg0ire: Same test without changing the visibility : https://3v4l.org/F0Xtu

greg0ire: And finally, a test with incompatible signatures : https://3v4l.org/G65su

Soullivaneuh: > This is perfectly logical if you see it from a visibility point of view: the child class know nothing about the parent's private scope, so it can override anything that's private, but can't access the parent.

In this case, it should not care of the parent syntax too, but it does:

https://3v4l.org/oCG5c

Possible override of a private method is a whole of possible errors for me...

But PHP allows it, right. :-/

In this case yes, `final` keyword make sense...

greg0ire: I think there is no point in making this `final` though : people can do whatever they want in child classes, they won't be able to call the parent class method.

sagikazarmark: Depends: if it makes sense to have custom id resolution logic, leave it as is. Otherwise I would make it final.

Soullivaneuh: I still think overriding private method is insane as possible. What is a goal of protected in this case?

But right, make it `final`.

greg0ire: Calling the method from the parent class : https://3v4l.org/RjFlU : the parent private method is called, not the "overriden" method. So there is no real override at all…

sagikazarmark: @greg0ire is right, although in this case I wonder why https://3v4l.org/oCG5c happens

OskarStark: > In this case yes, final keyword make sense...

👍

greg0ire: I just [asked this on SO](http://stackoverflow.com/questions/37792784/why-isnt-php-happy-with-a-public-method-signature-being-incorrect-with-the-pare), wait and see, maybe it's a bug ?

Block/Service/AbstractCategoriesBlockService.php
greg0ire: > Some support would be welcome if you want to chime in against this kind of nonsense…

To clarify because I might not have been very clear here, if any discussion should take place about this, it should take place [here](https://github.com/symfony/symfony/issues/18890) IMO.

Block/Service/AbstractCategoriesBlockService.php
core23: I don't think that it makes sense to make a private method final, so I leave it as it is:
 * nobody can call the method from the outside
 * nobody could extend a private method

=> Why the hell should we make it final? Nobody knows the method (from the outside). It's no API method!

Soullivaneuh: > nobody could extend a private method

https://3v4l.org/oCG5c

> Why the hell should we make it final? Nobody knows the method (from the outside). It's no API method!

https://3v4l.org/oCG5c

Yes, it's horrifying. But this is technically possible.

In this case, make it final would avoid future issue like "You broke my ugly code making an ugly override on the last release!".

core23:
* Added hu translation
* Added type check to `getCategory`, `getCollection`, `getTag`

Soullivaneuh:
@OskarStark please wait https://github.com/sonata-project/SonataClassificationBundle/pull/201#discussion_r66782389 before merge.

Block/Service/AbstractCategoriesBlockService.php
sagikazarmark: This is only some kind of partial/weird extension: https://3v4l.org/G65su
This is incompatible.

greg0ire: > This is only some kind of partial/weird extension: https://3v4l.org/G65su

It's not : it's too methods that are completely invisible to each other. Makes perfect sense to me…

sagikazarmark: I was reflecting to the previous comment by @Soullivaneuh . His examples shows that in some cases incompatibilities of inherited private methods cause error.

greg0ire: Oh. You linked to my example, that's why.

core23: > https://3v4l.org/oCG5c
> Yes, it's horrifying. But this is technically possible.

That's just an example, that you can't declare the same private method name, but you can't technically override it, like protected or public methods. BUT that's no reason to make it final. That doesn't solve the problem. The only thing that changes is the kind of compiler error.

sagikazarmark: I disagree. If you "override" the method and it does something nasty then you don't necessarily get an error, but mess up your database. If you want to change some behaviour, you can change the lifecycle events, but this method should never be changed/overridden, etc, thus it should be final.

Making it final NOW doesn't hurt, as making it non-final later. It can be a problem in the other case.

greg0ire: @sagikazarmark : I think [you can't override a private method to do something fucked up](https://3v4l.org/BdBAW).

OskarStark:
> @OskarStark please wait #201 (comment) before merge.

will do, the conversation started after my comment ;-)

Block/Service/AbstractCategoriesBlockService.php
sagikazarmark: Hm, good to know. And it is still scary how this thing works.

greg0ire: > Hm, good to know. And it is still scary how this thing works.

Yeah, not sure it won't suddenly change… people on SO don't know, and told me to ask this to internals.

greg0ire: So in conclusion I'd say don't make it final.

Soullivaneuh: > So in conclusion I'd say don't make it final, because override won't work, even if you are using php 7.

Am I missing something? This can be overridden according to previous examples.

greg0ire: > Am I missing something ?

Possibly. Let's have a look at the last example I gave :

```
<?php

class A {
    private function test() {
        echo 'a';
    }
    
    public function otherTest() {
        return $this->test();
    }
}

class B extends A {
    public function test() {
        echo 'something fucked up';
    }

}

$b = new B();
$b->otherTest();
```

this outputs "a". Which means you can do whatever the fuck you want in the child class, the parent does not care. You haven't overriden anything, and by "override", I mean make a function call in the parent execute a new logic defined in the child class. Defining a method with the same name is not "overriding", to me, in this case.

core23:
>>  @OskarStark please wait #201 (comment) before merge.

> will do, the conversation started after my comment ;-)

Thought we finished the discussion. IMO this is no blocker for this PR. 

IF we want to use hard limit, we should apply it automatically to all files.

core23:
@Soullivaneuh Anything to do here?

Block/Service/AbstractCategoriesBlockService.php
core23: Discussion continues here https://github.com/sonata-project/dev-kit/issues/185

core23:
Ping @Soullivaneuh

Soullivaneuh:
@core23 code looks good. I let another contributor knowing well this bundle merge this one.

OskarStark:
LGTM 👍 , what about https://github.com/sonata-project/SonataClassificationBundle/pull/201#discussion_r66782389 ??

Can we merge this?

Soullivaneuh:
Nothing else to do here.

OskarStark:
Thank you @core23 👍 

Great work 🍻

PR #203, opened 2016-05-26
Add PHP dependency

Th3Mouk:
### Changelog

Add PHP dependency on Composer.

```markdown
### Added
- Composer `PHP` dependency
```


Soullivaneuh:
Well, one commit would be enough but let's keep it as is. Thanks!

Issue #174, opened 2016-02-24
Property count does not exist


webdevilopers:
Yesterday my setup still seemed to work. Today I get:

```
  [RuntimeException]                                                          
  An error occurred when executing the "'cache:clear --no-warmup'" command:   
                                                                              
    [Doctrine\ODM\MongoDB\Mapping\MappingException]                           
    An error occurred in Sonata\ClassificationBundle\Document\BaseCategory    
                                                                              
                                                                              
    [ReflectionException]                                                     
    Property count does not exist                                                                                      
```

I havn't seen any commit on this bundle that could cause this.

But I recognized this commit by @OskarStark :
https://github.com/sonata-project/SonataAdminBundle/commit/926f15924f612c75366472d99bedf8c6ca25ea68

It relates to MongoDB Collection. Maybe this is linked to the `property count` issue?

```php
$objectCount = count($value);
$postCount   = count($data[$childFormBuilder->getName()]);
```
Changed to:
```php
$collection->add(new $entityClassName());
$propertyAccessor->setValue($entity, $path, $collection);
```

Composer:
```javascript
    "require": {
        "php": ">=5.3.3",
        "symfony/symfony": "2.6.*",
        "doctrine/orm": "2.5.*",
        "doctrine/dbal": "~2.4",
        "doctrine/doctrine-bundle": "~1.2",
        "doctrine/mongodb-odm": "^1.0",
        "doctrine/mongodb-odm-bundle": "~3.0",
        "sonata-project/core-bundle": "dev-master",
        "sonata-project/admin-bundle": "dev-master",
        "sonata-project/doctrine-orm-admin-bundle": "dev-master",
        "sonata-project/cache-bundle": "dev-master",
        "sonata-project/classification-bundle": "dev-master",
        "sonata-project/media-bundle": "dev-master#1cc799eda0fa7dc0fb01045cb36747e89b4bffcf",
        "sonata-project/datagrid-bundle": "dev-master",
        "sonata-project/user-bundle": "dev-master",
        "sonata-project/intl-bundle": "dev-master",
        "sonata-project/easy-extends-bundle": "dev-master",
        "sonata-project/doctrine-mongodb-admin-bundle": "dev-master",
        "sonata-project/translation-bundle": "dev-master",
```

webdevilopers:
I can see the property `count` inside the ODM Mapping of `Resources/config/doctrine/BaseCategory.mongodb.xml`:
```xml
[...]

PR #205, opened 2016-05-26
Fixed wrong mongodb properties

core23:
Closes https://github.com/sonata-project/SonataClassificationBundle/issues/174

### Changelog

```markdown
### Fixed
- Removed unmapped `count` property in `BaseCategory.mongodb.xml`
- Renamed wrong `slug` property to `name` in `BaseContext.mongodb.xml`
```

### Subject

Fixed wrong and unmapped mongodb properties.

### DIFF
--- a/Resources/config/doctrine/BaseCategory.mongodb.xml
+++ b/Resources/config/doctrine/BaseCategory.mongodb.xml
@@ -11,7 +11,6 @@
         <field name="slug"          type="string"       fieldName="slug"    />
 
         <field name="description"   type="string"   fieldName="description"    />
-        <field name="count"         type="int"      fieldName="count"          />
 
         <field name="createdAt"    type="timestamp"   fieldName="createdAt" />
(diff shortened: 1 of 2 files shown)


OskarStark:
If they are already generated, it is a BC break, isn't it?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [3.x]
+### Fixed
+ - Removed unmapped `count` property in `BaseCategory.mongodb.xml`
```
Soullivaneuh: Should be on PR body only.

core23:
In general I would say yes, it is a BC break. But AFAIK you can't use mongodb with this bundle, because there are no properties in the php classes. 

This PR makes mongodb useable again.

OskarStark:
> This PR makes mongodb useable again.

then we should merge this as it is 👍

PR #221, opened 2016-06-09
Add documentation for admin 

clytemnestra:
### Subject

 Added documentation about how to remove Admin, maybe add more info about it in the future. We need an explanation about the admin interface and how to remove it, if people don't want additional fields in their admin dashboard.

### To do

- [ ] wait for opinions
- [ ] squash commits

### DIFF
--- a/Resources/doc/reference/admin.rst
+++ b/Resources/doc/reference/admin.rst
@@ -1,0 +1,52 @@
+Admin
+=====
+
+The bundle comes with an Admin interface for managing your contexts, tags, collections and categories.
+
+Removing the Admin 
+------------------
+
+There may be cases where you don't want the default admins to appear on your dashboard. For that, you need to set up a compiler pass that tags the services with ``show_in_dashboard = false``:
+
+
+.. code-block:: php
+
+  use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+  use Symfony\Component\DependencyInjection\ContainerBuilder;
+  use Symfony\Component\DependencyInjection\Definition;
+  
+  class CompilerPass implements CompilerPassInterface
+  {
+      public function process(ContainerBuilder $container)
+      {
+
+          $definitionsNames = array('sonata.classification.admin.category', 'sonata.classification.admin.tag', 'sonata.classification.admin.context', 'sonata.classification.admin.collection');
+  
+          foreach ($definitionsNames as $definitionName) {
+  
+              /** @var Definition $definition */
+              $definition = $container->getDefinition($definitionName);
+  
+              $tags = $definition->getTags();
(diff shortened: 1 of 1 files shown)


Resources/doc/reference/admin.rst
```diff
@@ -0,0 +1,53 @@
+Admin
+=======
+
+The bundle comes with an Admin interface for managing your contexts, tags, collections and categories.
+
```
Soullivaneuh: Remove one line.

Resources/doc/reference/admin.rst
```diff
@@ -0,0 +1,53 @@
+Admin
+=======
```
Soullivaneuh: Should be same length as the title.

Soullivaneuh:
Please change you commit title.

"Create admin.rst" means nothing clear.

core23:
Maybe we should move the docs to the AdminBundle instead of copy & paste it for every bundle

clytemnestra:
The thing is sometimes the admin services are required and you can't remove them from the container without having to override the places where they are used. And that is specific to each bundle.

Resources/doc/reference/admin.rst
```diff
@@ -0,0 +1,52 @@
...
+              $tags['sonata.admin'][0]['show_in_dashboard'] = false;
+              $definition->setTags($tags);
+  
+          }
+      }
+  }
+
+
+
+Note that this will keep the services and the routes in the container, it will just not show links in the dashboard, anymore.
```
greg0ire: "dashboard anymore"

Resources/doc/reference/admin.rst
```diff
@@ -0,0 +1,52 @@
...
+
+If you don't want the admin routes to be accessible, either, then you will have to remove the service definitions from the container.
```
greg0ire: remove extra comma

clytemnestra:
Squashed and changed

core23:
I don't like the idea to create a whole class in the docs which removes the automapping to the admin. Maybe we should find another way to include this feature in the bundle. Maybe via a config key?

Same as https://github.com/sonata-project/SonataMediaBundle/pull/1048

clytemnestra:
That could work, too. But then you'd have to add a conf. key for every admin service, in case one wants to remove just some of them.

clytemnestra:
Well I guess if you'd want one removed you would want them all and there's always this alternative in case you wanted only a few removed.

core23:
IMO, we don't need any special information for this rare case.

You could also configure your admins in the [configuration](https://sonata-project.org/bundles/admin/3-x/doc/reference/dashboard.html#using-the-config-yml).

PR #196, opened 2016-05-26
Remove media

core23:
Adding the `SonataMediaBundle` as a dependency was't a good idea, because we have a cross-dependency between both bundles. This PR removes this dependency completly.

The dependency was needed for a category and collection thumb, which should be handled in a custom `Application\SonataClassificationBundle` implementation.

PR #224, opened 2016-06-17
Remove upper limit for media-bundle

core23:
I am targetting this branch, because the build in the master branch is broken.

## Changelog

```markdown
### Removed
- removed upper conflict limit for `sonata-project/media-bundle` dependency
```

## Subject

This should fix the build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "conflict": {
         "jms/serializer": "<0.13",
-        "sonata-project/media-bundle": "<3.0 || >=5.0"
+        "sonata-project/media-bundle": "<3.0"
     },
     "autoload": {
         "psr-4": { "Sonata\\ClassificationBundle\\": "" }


Soullivaneuh:
But I don't think this is the good way. Removing the high limit is dangerous.

BTW, why this solve the issue (didn't get time to dig on it...)?

core23:
> BTW, why this solve the issue (didn't get time to dig on it...)?

Haven't figured out yet. 

Maybe we can merge this to have a good base for new tickets on the master branch and create a new ticket to dig deeper and fix the upper limit later. There is no reason for the upper limit at the moment.

Soullivaneuh:
> There is no reason for the upper limit at the moment.

Wrong. The upper limit is her to prevent new major dependencies break.

Sorry but we won't merge a MR juste because "I don't know why but it works if I do." :wink:

core23:
IMHO requiring the `dev-master` branch is interpreted as `>=5.0`

core23:
@greg0ire @OskarStark Do you have any idea why this solves the problem or if there is a better solution?

Soullivaneuh:
> IMHO requiring the `dev-master` branch is interpreted as `>=5.0`

Impossible: https://packagist.org/packages/sonata-project/media-bundle#dev-master

Maybe something to do on media-bundle itself?

See also what I done for admin and block bundles.

greg0ire:
> @greg0ire @OskarStark Do you have any idea why this solves the problem or if there is a better solution?

I see at least two problems here : 

- we are trying to install 2 master versions together, which should never happen, because we shouldn't develop on sand, but on rock-solid code. This is because having a dev-master dependency means you cannot release a stable version, because this dependency can create a BC-break at all times for the user (who might be using it, who knows?)
- Sonata Media 3 requires classification, which means if there is integration to be done, it should be done on the media bundle, otherwise we get a circular dependency. Or maybe it should be the other way around? Not both, that's for sure.

So to sum up, we should have a look at why this circular dependency exists, and fix it by moving all the integration on one side, or move it to another package, or create abstraction in another package and only rely on these abstractions.

core23:
Maybe we can solve this after merging: https://github.com/sonata-project/SonataClassificationBundle/pull/196

Soullivaneuh:
> See also what I done for admin and block bundles.

@core23 Did you take a look?

core23:
Got some news from the composer crew:

> Yup the problem seems to be that for requires, we allow either dev-master OR the alias to work, but for conflict as well it conflicts with either dev-master OR the alias, while for conflict it really only shouldn't conflict if the alias allows it, but I'm not sure this is solvable. Aliases are a convenience but they are also a hack to some extent and making them work in all edge cases might not be feasible reasonably.

> Maybe you can drop the >=5 conflict, or turn it into a >=5 && <99, that should solve it I assume.

Soullivaneuh:
The upper limit is here for a preventive way, this can't be removed.

Also:

>> See also what I done for admin and block bundles.

> @core23 Did you take a look?

No time to find the related PR right now, but AFAIK, it was the same issue.

core23:
>> See also what I done for admin and block bundles.

> @core23 Did you take a look?

Both master versions of the bundles (admin and block) are loading the latest stable (3.x) version:
https://travis-ci.org/sonata-project/SonataBlockBundle/jobs/144193943#L407
https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/144035597#L420

Your rule `^4.x@dev` rule doesn't apply, because only stable versions are allowed.

core23:
@Soullivaneuh Maybe we can accept this for the master branch, because we should drop the media-bundle soon on the `master` branch. See https://github.com/sonata-project/SonataClassificationBundle/pull/196

The other sultion is adding an ugly workaround, as mentioned by the composer guys.

>> Thank you for you analysis. Dropping the conflict rule was my first idea to solve this issue temporary. 
The question is, if this is a bug or `works as designed`. Are you going to fix this?

> As I hinted at, I am not sure that we can solve it, at least within a reasonable amount of time. So I'd rather you use a workaround like >=5 && <999 which is safe enough, if it works :)

Soullivaneuh:
> Your rule `^4.x@dev` rule doesn't apply, because only stable versions are
allowed.

Wrong. `@dev` notation bypasses the default rule. :-)

Could you please try that?

Sullivan SENECHAL

Le 14 juil. 2016 17:02, "Christian Gripp" <notifications@github.com> a
écrit :

core23:
Have a look at the travis build. It doesn't apply :(

core23:
@OskarStark @greg0ire What do you think?

greg0ire:
> @OskarStark @greg0ire What do you think?

TBH, I think I don't understand the problem anymore, it's too complicated for me :P
As I said [here](https://github.com/sonata-project/SonataClassificationBundle/pull/224#issuecomment-228866096), this would be a lot more simple if we stopped referencing unstable branches from other bundles.

core23:
> As I said here, this would be a lot more simple if we stopped referencing unstable branches from other bundles.

I agree with that, but we have a circular dependency between the classification and the media bundle in the master branch.

As I said, we should do something to get a green build for this master branch and remove the circular dependency afterwards.

core23:
Ping @Soullivaneuh

core23:
Closing this in favor of https://github.com/sonata-project/SonataClassificationBundle/pull/196

PR #229, opened 2016-07-28
using cocur-slugify for tag slugs

Vesax:
I am targetting this branch, because this is a BC fix

## Changelog

```markdown
### Fixed
- incorrect names transliterating for slugs
```

## Subject

I found a problem with generating slug for russian letters.

Code

```
if (function_exists('iconv')) {
    $text = iconv('utf-8', 'us-ascii//TRANSLIT', $text);
}
```

doesn't work correctly, slugify method returns n-a for all russian tags and it makes a problem with unique constraint.

Actually, we have the cocur/slugify package required in core-bundle, so, we can just use it for generating slugs in classification bundle. It fixes the problem with transliterating.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     ],
     "require": {
         "php": "^5.3 || ^7.0",
+        "cocur/slugify": "^1.4 || ^2.0",
         "sonata-project/admin-bundle": "^3.0",
         "sonata-project/datagrid-bundle": "^2.2",
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
--- a/Model/Tag.php
+++ b/Model/Tag.php
@@ -11,6 +11,8 @@
 
 namespace Sonata\ClassificationBundle\Model;
 
+use Cocur\Slugify\Slugify;
+
 abstract class Tag implements TagInterface
 {
     /**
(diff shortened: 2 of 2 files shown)


greg0ire:
> Actually, we have the cocur/slugify package required in core-bundle, so, we can just use it…

No we can't, if you want to use it here you have to explicitely require it in the classification-bundle too. That way, if it gets removed from the core bundle, this piece of code will not be impacted. This will also make slugify upgrades smoother : we will not forget to upgrade the code in this bundle too.

Vesax:
@greg0ire, thanks, you are right. I have required this library in classification-bundle too

greg0ire:
LGTM!

composer.json
```diff
@@ -27,7 +27,8 @@
         "symfony/options-resolver": "^2.3 || ^3.0",
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
         "sonata-project/admin-bundle": "^3.0",
-        "sonata-project/datagrid-bundle": "^2.2"
+        "sonata-project/datagrid-bundle": "^2.2",
+        "cocur/slugify": "^1.4 || ^2.0"
```
core23: Please sort dependencies, move this to the top

greg0ire: I almost asked this, but since dependencies are not already sorted, I thought it would be a bit noisy in the diff.

greg0ire: How about merging #230 first, then ask @Vesax to rebase?

Vesax: This is good idea

greg0ire: Thanks! Sadly, my build is stuck, I emailed Travis about that, but no answer so far.

greg0ire: Managed to trigger the build by force pushing :)

greg0ire: Passing and merged :) @Vesax , can you please rebase? You should get a conflict, fix it by adding the package at the right place please.

Vesax:
@SonataCI, yes sir!

greg0ire:
Looks Good To Squash-Merge

greg0ire:
Looks like the build is stuck, like mine was :(

Can you : 

1. Use `git rebase --interactive` to fixup the second commit into the first one?
2. Use `git push --force` to replace the 2 commits with the brand new commit obtained at step 1? It could trigger the build.

Vesax:
Finally, seems like everything is okay

greg0ire:
Thanks @Vesax !

OskarStark:
Good job @Vesax 👍

OskarStark:
`patch` or `minor` @greg0ire ?

greg0ire:
I'd say patch : we are fixing something.

PR #232, opened 2016-09-03
Fix old Admin class usage

binhle410:
I am targetting this branch to fix "**PHP Fatal error:  Class 'Sonata\AdminBundle\Admin\Admin' not found**"

## Changelog

```markdown
### Changed
- Use `AbstractAdmin` instead of deprecated `Admin`
```

### DIFF
--- a/Tests/Admin/AdminTest.php
+++ b/Tests/Admin/AdminTest.php
@@ -11,7 +11,6 @@
 
 namespace Sonata\ClassificationBundle\Tests\Admin;
 
-use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\ClassificationBundle\Model\ContextManagerInterface;
 
 class AdminTest extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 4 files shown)


core23:
I switchted the base of your PR, because this should be fixed for the stable branch first.

core23:
Please have a look at the `ContextAdmin` too and change the `composer.json` dependency for the admin bundle.

Soullivaneuh:
Please write a changelog.

Soullivaneuh:
Also, some test to prove the bug would be great.

core23:
How would you create a test for this @Soullivaneuh ?

greg0ire:
Instantiating a child class of `ContextAwareAdmin` would do the trick, wouldn't it? Or maybe even mocking it would work?

binhle410:
core23 has already rebased, so I don't need to do it anymore correct?
The only thing left for this change to be merged is to provide a PHPUnit test ?

greg0ire:
Yes and yes.

binhle410:
I have added a Test for it under AdminTest class, please have a look.

Tests/Admin/AdminTest.php
```diff
@@ -25,6 +26,18 @@ public function setUp()
         $this->contextManager = $this->getMock('Sonata\ClassificationBundle\Model\ContextManagerInterface');
     }
 
+    public function testAbstractAdminChildren()
+    {
+        $contextAwareAdmin = $this->getMockBuilder('Sonata\ClassificationBundle\Admin\ContextAwareAdmin')
+            ->disableOriginalConstructor()
+            ->getMock();
+        $this->assertInstanceOf(AbstractAdmin::class, $contextAwareAdmin);
```
greg0ire: This will not work with old versions of php (as Travis shows). Otherwise, looks good. It will not fail if the old class is extended, but a deprecation error will be triggered, which should show up in the `SYMFONY_DEPRECATIONS_HELPER=0` build, so we're all good

binhle410: So, how old would the php version be for this test to not work?

greg0ire:
You must require `sonata/admin-bundle` `^3.1` to fix the build.

greg0ire:
LGTM

greg0ire:
ping @sonata-project/contributors

PR #236, opened 2016-09-19
fix context foreign key constraint

fullpipe:
# Changelog
```markdown
### Fixed
 - Set context to null on delete
```
# Subject
When purging database with doctrine fixtures bundle. You receive error
```
An exception occurred while executing 'DELETE FROM classification__context':                                                                                             
  SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`wikium`.`classification__category`, CONSTR  
  AINT `FK_43629B36E25D857E` FOREIGN KEY (`context`) REFERENCES `classification__context` (`id`))
```

### DIFF
--- a/DependencyInjection/SonataClassificationExtension.php
+++ b/DependencyInjection/SonataClassificationExtension.php
@@ -158,7 +158,7 @@ class SonataClassificationExtension extends Extension
                 array(
                     'name' => 'context',
                     'referencedColumnName' => 'id',
-                    'onDelete' => 'set null',
+                    'onDelete' => 'SET NULL',
                 ),
             ),
             'orphanRemoval' => false,
--- a/DependencyInjection/SonataClassificationExtension.php
+++ b/DependencyInjection/SonataClassificationExtension.php
@@ -158,6 +158,7 @@ class SonataClassificationExtension extends Extension
                 array(
                     'name' => 'context',
                     'referencedColumnName' => 'id',
+                    'onDelete' => 'set null',
                 ),
             ),
             'orphanRemoval' => false,


DependencyInjection/SonataClassificationExtension.php
```diff
@@ -158,6 +158,7 @@ public function registerDoctrineMapping(array $config)
                 array(
                     'name' => 'context',
                     'referencedColumnName' => 'id',
+                    'onDelete' => 'set null',
```
greg0ire: I would use capitals here, so that people understand this is SQL

greg0ire:
Could this be a bug in the fixtures bundle? I mean it should be able to delete the tables in the right order, right?

DependencyInjection/SonataClassificationExtension.php
greg0ire: Thanks for changing that.

greg0ire:
I had a look and I feel version 1.2 of the library is buggy. Can you try with version 1.1 ? If 1.1 works, I think the right fix would be to blacklist 1.2.0 and 1.2.1 in Composer.

fullpipe:
Yes, you right. I found this [PR](https://github.com/doctrine/data-fixtures/pull/239) which should solve this problem.

Using `"doctrine/data-fixtures": "1.1.1"` is working solution.

Thanks

greg0ire:
> If 1.1 works, I think the right fix would be to blacklist 1.2.0 and 1.2.1 in Composer.

Can you reopen this PR and force push a solution doing this instead then?

fullpipe:
But `sonata-project/SonataClassificationBundle` does not use `doctrine/data-fixtures`. Or your meen to make PR to `doctrine/DoctrineFixturesBundle`?

greg0ire:
> But sonata-project/SonataClassificationBundle does not use doctrine/data-fixtures. Or your meen to make PR to doctrine/DoctrineFixturesBundle?

Oh no you're right I overlooked that. Please disregard my comment.

PR #242, opened 2016-11-01
fix-context command throws a fatal error. Incorrect variable is referenced during execution. 

jayesbe:
I am targeting this branch, because this issue exists on this branch. 

## Changelog

```markdown
### Changed
- `tag` variable reference inside collection iterator to point to `collection` variable. The incorrect reference was throwing an error when executing sonata:classification:fix-context from the command line.
```

## Subject

Collection iterator references tag variable and should be using collection.


jayesbe:
@greg0ire Id like to.. it's just too much work for something that is just broken and is a 10 second fix. Unless you can suggest how to make it more convincing.

greg0ire:
I'm on my phone and now I have to try to look at this file and understand your changes. Github shows me very little context in the diff, so this is not very convenient. You changed the name of a variable twice, which is honestly not that big of a change. You should have some energy left to produce a precise convincing message like this one :

```
Reference the correct variable 

It looks like this code block was copied from another code block and the author forgot to replace every reference of the tag variable. 
```

Again, I'm reviewing this on my phone so maybe I did not understand the problem. I should be able to just by reading the commit message. If you have a look at the history of this file, you will see someone else has already use `update fixcontextcommand`. Let's try to improve on that, OK? Otherwise the history will become useless.

jayesbe:
@greg0ire understood. Will update when I get the chance.

greg0ire:
Great! Thanks a lot!

jayesbe:
@greg0ire I updated the commit. Is it ok now?

greg0ire:
You updated the PR title. The commit itself did not change.

jayesbe:
mangled the repo. pure waste of time on this ticket.

new PR #243

PR #243, opened 2016-11-09
fix-context command throws a fatal error.

jayesbe:
I am targeting this branch, because this issue exists on this branch. 

## Changelog

```markdown
### Changed
- `tag` variable reference inside collection iterator to point to `collection` variable. The incorrect reference was throwing an error when executing sonata:classification:fix-context from the command line.
```

## Subject

Collection iterator references tag variable and should be using collection.


Soullivaneuh:
The changelog note should be on the `fixed` section.

PR #246, opened 2016-11-14
French translations + menu icon.

jlamur:
I am targetting this branch, because it will not break BC.

## Changelog

```markdown
### Added
- Menu icon (fa-tags).
- Fix bad / missing translations in french.
```

## Subject
- choose `fa-tags` instead of `fa-bars` used everywhere for this Bundle (in the documentation at least),
- harmonize breadcrumbs `_list` with simply the name of the entity instead of using "List of XXX", because it is the most common usage in other bundles (idem for `_create`'s breadcrumbs).

### DIFF
--- a/Resources/translations/SonataClassificationBundle.fr.xliff
+++ b/Resources/translations/SonataClassificationBundle.fr.xliff
@@ -28,7 +28,7 @@
             </trans-unit>
             <trans-unit id="breadcrumb.link_category_tree">
                 <source>breadcrumb.link_category_tree</source>
-                <target>Catégorie</target>
+                <target>Catégories</target>
             </trans-unit>
             <trans-unit id="list.label_name">
                 <source>list.label_name</source>
@@ -68,7 +68,7 @@
             </trans-unit>
             <trans-unit id="breadcrumb.link_tag_list">
                 <source>breadcrumb.link_tag_list</source>
-                <target>Liste des tags</target>
+                <target>Tags</target>
             </trans-unit>
             <trans-unit id="form.label_slug">
                 <source>form.label_slug</source>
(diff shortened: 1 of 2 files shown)


core23:
You have done three different things in this PR. Please splitt

jlamur:
Would you like me to split my commits, split my PR or both ?

I can even make 4 commits && PR if it is better !

It is the first time I create a PR, so I'm sorry if my question seems dumb, I don't really know what all of that implies to you.

jlamur:
My commits are split. I will create new PR for each of them if you confirm me that it is what I am supposed to do.

Resources/translations/SonataClassificationBundle.fr.xliff
```diff
@@ -12,7 +12,7 @@
             </trans-unit>
             <trans-unit id="label_tags">
                 <source>label_tags</source>
-                <target>Tags</target>
+                <target>Étiquettes</target>
```
greg0ire: I know "étiquette" is french and "tag" isn't, but in this context I think french people are more used to seeing "tag". This would maybe make sense in an `fr_CA` traduction, along with "courriel" or "patin à roulettes".

jlamur: @greg0ire You're right :) I dropped this commit.

greg0ire:
Split commits are enough.

OskarStark:
Please check travis @jlamur

jlamur:
@OskarStark tests are **not** failing because of my PR.
They pass with `--prefer-lowest` so I guess a BC break was introduced in a dependency's new version.

I will investigate and come back with a PR, if I can.

jlamur:
@OskarStark I'm not enough skilled to fix the bug, but I found the source of the problem, and I opened an issue for it: #277

jordisala1991:
LGTM 👍

greg0ire:
Thanks @jlamur !

OskarStark:
Thank you guys 👍

Issue #234, opened 2016-09-09
The options "category", "context" do not exist.


vub:
#### Sonata packages

``` bash
$ composer show sonata-project/*

sonata-project/admin-bundle                  3.7.0              The missing Symfony Admin Generator
sonata-project/block-bundle                  3.1.1              Symfony SonataBlockBundle
sonata-project/cache                         1.0.7              Cache library
sonata-project/classification-bundle         dev-master 5e9af17 Symfony SonataClassificationBundle
sonata-project/core-bundle                   3.1.1              Symfony SonataCoreBundle
sonata-project/datagrid-bundle               2.2                Symfony SonataDatagridBundle
sonata-project/doctrine-mongodb-admin-bundle 3.0.0              Symfony Sonata / Integrate Doctrine MongoDB ODM into the SonataAdminBundle
sonata-project/doctrine-orm-admin-bundle     3.0.5              Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle           2.1.10             Symfony SonataEasyExtendsBundle
sonata-project/exporter                      1.7.0              Lightweight Exporter library
```
#### Symfony packages

``` bash
$ composer show symfony/*
symfony/monolog-bundle     2.11.1  Symfony MonologBundle
symfony/phpunit-bridge     v3.1.4  Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.2.0  Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-intl-icu  v1.2.0  Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring  v1.2.0  Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.2.0  Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php70     v1.2.0  Symfony polyfill backporting some PHP 7.0+ features to lower PHP versions
symfony/polyfill-util      v1.2.0  Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.0  Symfony Security Component - ACL (Access Control List)
symfony/swiftmailer-bundle v2.3.11 Symfony SwiftmailerBundle
symfony/symfony            v3.1.4  The Symfony PHP framework
```
#### PHP version

``` bash
$ php -v
PHP 5.6.23 (cli) (built: Jun 22 2016 12:15:20)
[...]

PR #249, opened 2016-11-25
SF3 compatibility

nicodmf:
All is in the title

### DIFF
--- a/Admin/CategoryAdmin.php
+++ b/Admin/CategoryAdmin.php
@@ -19,7 +19,6 @@ use Sonata\AdminBundle\Route\RouteCollection;
 class CategoryAdmin extends ContextAwareAdmin
 {
     protected $formOptions = array(
-        'cascade_validation' => true,
     );
 
     /**
@@ -47,11 +46,9 @@ class CategoryAdmin extends ContextAwareAdmin
             if ($this->getSubject()->getParent() !== null || $this->getSubject()->getId() === null) { // root category cannot have a parent
                 $formMapper
                   ->add('parent', 'sonata_category_selector', array(
-                      'category' => $this->getSubject() ?: null,
                       'model_manager' => $this->getModelManager(),
                       'class' => $this->getClass(),
                       'required' => true,
-                      'context' => $this->getSubject()->getContext(),
                     ));
             }
         }
--- a/Admin/CollectionAdmin.php
+++ b/Admin/CollectionAdmin.php
@@ -18,7 +18,6 @@ use Sonata\AdminBundle\Form\FormMapper;
 class CollectionAdmin extends ContextAwareAdmin
 {
     protected $formOptions = array(
-        'cascade_validation' => true,
     );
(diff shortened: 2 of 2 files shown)


greg0ire:
Please do not ignore the PR template.

Admin/CategoryAdmin.php
```diff
@@ -47,11 +46,9 @@ protected function configureFormFields(FormMapper $formMapper)
             if ($this->getSubject()->getParent() !== null || $this->getSubject()->getId() === null) { // root category cannot have a parent
                 $formMapper
                   ->add('parent', 'sonata_category_selector', array(
-                      'category' => $this->getSubject() ?: null,
                       'model_manager' => $this->getModelManager(),
                       'class' => $this->getClass(),
                       'required' => true,
-                      'context' => $this->getSubject()->getContext(),
```
greg0ire: This is not the good solution 👎 , see #234

greg0ire: [Here is how you can fix this](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Form/Type/AdminType.php#L108)

Admin/CategoryAdmin.php
```diff
@@ -19,7 +19,6 @@
 class CategoryAdmin extends ContextAwareAdmin
 {
     protected $formOptions = array(
-        'cascade_validation' => true,
```
greg0ire: Can you please explain why you are doing this?

nicodmf: Because this create a bug

nicodmf:
Personnally i don't know if there solution are good or not, but as it the bundle is just a pain which take my evening.

Admin/CategoryAdmin.php
greg0ire: Can you elaborate?

greg0ire:
> Personnally i don't know if there solution are good or not, but as it the bundle is just a pain which take my evening.

We're very grateful for your evening. If you want to fix this, I will help you. But it has to be done so that it works for sf 2.3 too.

greg0ire:
> Could you please provide a better compatibility and stability ?

I think there are not many bugs to resolve, but we need someone to provide it. I shouldn't do it because I don't use this bundle, but if you want to do it, you'll have fast feedback from me, and I will help you make this mergeable.

Admin/CategoryAdmin.php
greg0ire: Found the answer myself : https://github.com/webmozart/symfony/blob/d85c8173bbc2ff49186d3be7ed9e4c00c3a18bf9/UPGRADE-2.6.md#form It is deprecated since 2.6

greg0ire: validation.xml should be changed accordingly.

greg0ire:
Don't bother, I did it myself, see #250 . Thanks for trying to help!

PR #256, opened 2016-12-22
fixed fosrest 2.0 support

lemorragia:
Somehow this one slipped away in fosrest 2.0 support. Like in sonata-user and sonata-media, serializationGroups need to be arrays and not strings to support both fos-rest 1 and 2.

## Changelog

```markdown
### Changed
- Changed api controller serializationGroups from strings to arrays
```
## Subject

This should complete fos-rest >= 2.0 support for this bundle too. Its bc-compatible.


OskarStark:
Thank you @lemorragia 

I modified the commit message a bit:
`Use array instead string for serialization groups to support FosRest 2.0`

PR #261, opened 2016-12-30
Add context field validation in tag/collection forms

dmarkowicz:
I am targetting this branch, because it is BC.

## Changelog

```markdown
### Added
- context fields validation in tag/collection create forms, which prevents creating objects
  with empty context from UI
```

## Subject
Current create forms allow user to create objects with empty context, which is a bug (`sonata:classification:fix-context` command try to repair such situations).

This PR
- add `required` attribute and `NotBlank` constraint to tag/collection forms to show nice message to user.
- prevent creating tag/collection with no context from UI

### DIFF
--- a/Resources/config/validation.xml
+++ b/Resources/config/validation.xml
@@ -19,6 +19,9 @@
                 <option name="max">32</option>
             </constraint>
         </property>
+        <property name="context">
+            <constraint name="NotBlank"/>
+        </property>
     </class>
     <class name="Sonata\ClassificationBundle\Model\Collection">
         <property name="name">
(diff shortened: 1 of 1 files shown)


core23:
👎 For adding the context field to the edit page(s). You could create a parent child connection with different contexts which could cause problems. 

Please use the `getNewInstance` or `pre*` methods instead.

dmarkowicz:
@core23 you're right, I forgot that context in tree should be the same. Removed category part from this PR for now. I will discuss category UI changes in another Issue.

Admin/CollectionAdmin.php
```diff
@@ -53,7 +54,12 @@ protected function configureFormFields(FormMapper $formMapper)
             ->add('description', 'textarea', array(
                 'required' => false,
             ))
-            ->add('context')
+            ->add('context', null, array(
+                'constraints' => array(
```
core23: Can we move this stuff to the `validation.xml` file?

dmarkowicz: Done.

OskarStark:
Thank you @dmarkowicz and sorry for the delay 👍

PR #270, opened 2017-01-20
Symfony3.*: use ArrayChoiceList instead of deprecated SimpleChoiceList

kindratmakc:
I am targetting this branch, because PR isn't break BC.

Changelog

```markdown
### Fixed
- Fix deprecated `SimpleChoiceList`
```

### DIFF
--- a/Form/Type/CategorySelectorType.php
+++ b/Form/Type/CategorySelectorType.php
@@ -15,6 +15,7 @@ use Sonata\ClassificationBundle\Model\CategoryInterface;
 use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
 use Sonata\CoreBundle\Model\ManagerInterface;
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\ChoiceList\ArrayChoiceList;
 use Symfony\Component\Form\Extension\Core\ChoiceList\SimpleChoiceList;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
@@ -50,6 +51,11 @@ class CategorySelectorType extends AbstractType
         $this->configureOptions($resolver);
     }
 
+    /**
+     * NEXT_MAJOR: replace usage of deprecated SimpleChoiceList.
+     *
+     * {@inheritdoc}
+     */
     public function configureOptions(OptionsResolver $resolver)
     {
         $that = $this;
@@ -58,7 +64,11 @@ class CategorySelectorType extends AbstractType
             'context' => null,
             'category' => null,
             'choice_list' => function (Options $opts, $previousValue) use ($that) {
-                return new SimpleChoiceList($that->getChoices($opts));
+                if (class_exists('Symfony\Component\Form\ChoiceList\ArrayChoiceList')) {
+                    return new ArrayChoiceList($that->getChoices($opts));
+                }
+
+                return new SimpleChoiceList($that->getChoices($opts));      // BC with Symfony 2.*
             },
         ));
     }
--- a/Tests/Form/Type/CategorySelectorTypeTest.php
+++ b/Tests/Form/Type/CategorySelectorTypeTest.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Tests\Form\Type;
+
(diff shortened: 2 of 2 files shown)


Form/Type/CategorySelectorType.php
```diff
@@ -58,7 +64,11 @@ public function configureOptions(OptionsResolver $resolver)
             'context' => null,
             'category' => null,
             'choice_list' => function (Options $opts, $previousValue) use ($that) {
-                return new SimpleChoiceList($that->getChoices($opts));
+                if (class_exists('Symfony\Component\Form\Extension\Core\ChoiceList\SimpleChoiceList')) {    // BC with Symfony 2.*
```
greg0ire: Rather than using old things while still available, how about using new things as soon as available?

kindratmakc: it makes sense

greg0ire:
Don't try to merge the solution I suggested with the previous solution : just overwrite it.

1. revert to it, undoing the merge in the process : `git reset --hard 62b57b4`
2. overwrite what is on github : `git push --force`

Tests/Form/Type/CategorySelectorTypeTest.php
```diff
@@ -0,0 +1,25 @@
...
+use Sonata\ClassificationBundle\Form\Type\CategorySelectorType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class CategorySelectorTypeTest extends \PHPUnit_Framework_TestCase
+{
+    public function testConfigureOptions()
+    {
+        $manager = $this->getMock('Sonata\CoreBundle\Model\ManagerInterface');
+        $categorySelectorType = new CategorySelectorType($manager);
+        $categorySelectorType->configureOptions(new OptionsResolver());
```
greg0ire: A test without assertions looks funny, but it's better to run this piece of code to ensure all the classes in use do exist, so good job!

kindratmakc: thx, it my first PR to open source project)

greg0ire: I noticed that, good job!

OskarStark: couldn't we add an assertion here? @greg0ire

greg0ire: We sure can… do you have one in mind?

ZRazor:
I'm new here, so dont judge, if I'm wrong, but this patch wont work as expected on symfony >=3.0
I applied this solution in my project on symfony 3.1 and noticed that category tree doesnt have any "-" symbols for child categories, actually it is not a tree anymore.
'choice_list' option is deprecated in 2.7 (according to documentation http://symfony.com/doc/2.7/reference/forms/types/choice.html )
and is not listed in 3.0 and above ( http://symfony.com/doc/3.0/reference/forms/types/choice.html )
I debugged a little to find out, why it doesnt throw exception:
Symfony\Component\Form\Extension\Core\Type::buildForm
In 2.7 it goes staight to this line:
```
$builder->addViewTransformer(new ChoiceToValueTransformer($options['choice_list']));
``` 
And in 3.0:
```
        $choiceList = $this->createChoiceList($options);
        $builder->setAttribute('choice_list', $choiceList);
```
And createChoiceList returns choices from choices option or from choice_loader, which is created by default. And then overrides anything we have passed to choice_list.
The solution for symfony 3.0 and above is to use 'choice_loader'. Need to pass there a class implementing ( http://api.symfony.com/3.1/Symfony/Component/Form/ChoiceList/Loader/ChoiceLoaderInterface.html )
In symfony 3.2 there is a new class CallbackChoiceLoader
```
    public function configureOptions(OptionsResolver $resolver)
    {
        $that = $this;
        $resolver->setDefaults(array(
            'context' => null,
            'category' => null,

            'choice_loader' => new CallbackChoiceLoader(function() use ($that) {
                return array_flip($that->getChoices());
            })
        ));
    }

```
Also I had to remove Options param from getChoices and childWalker functions, not sure if this is a good solution, but i checked - it never contains category option.

greg0ire:
Would you like to provide a PR @ZRazor ?

ZRazor:
@greg0ire, of course. I'try to make it compatible with 2.7 and >=3.0

greg0ire:
Sorry, but we won't merge that unless it's compatible with 2.3 :P … until the next major release, but I'm afraid it's not coming very soon.

ZRazor:
@greg0ire, i get it. I'll make sure that older versions work fine (with SimpleChoiceList)

greg0ire:
@ZRazor awesome!

OskarStark:
Closing in favour of: https://github.com/sonata-project/SonataClassificationBundle/pull/280#issuecomment-282038580

PR #274, opened 2017-02-02
3.3.0

qsomazzi:
Creation of a new release

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,9 +1,9 @@
 UPGRADE 3.x
 ===========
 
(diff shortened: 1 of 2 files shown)


UPGRADE-3.x.md
```diff
@@ -4,6 +4,9 @@ UPGRADE 3.x
 UPGRADE FROM 3.1 to 3.2
 =======================
 
+UPGRADE FROM 3.2 to 3.3
+=======================
+
```
greg0ire: Shouldn't this be above? Also, it's empty, so… not really needed IMO

PR #283, opened 2017-02-24
Make classification blocks more robust

core23:
I am targetting this branch, because this is a BC bugfix.

```markdown
### Fixed
 - Allow false value for category in `AbstractCategoriesBlockService`
 - Allow false value for collection in `AbstractCollectionsBlockService`
 - Allow false value for tag in `AbstractTagsBlockService`
```

## Subject

There is a problem in all `Abstract*TYPE*BlockService`s. E.g. when using the `AbstractCategoriesBlockService` without any special block setting the following exception occurs:

```
Argument 2 passed to Sonata\ClassificationBundle\Block\Service\AbstractCategoriesBlockService::get*() 
must implement interface Sonata\ClassificationBundle\Model\CategoryInterface, boolean given
```

The default option is a boolean value which causes a method mismatch. This PR changes the method signature of the (final) methods to allow any value (e.g. false) and checks if this a valid object for the block service.

### DIFF
--- a/Block/Service/AbstractCategoriesBlockService.php
+++ b/Block/Service/AbstractCategoriesBlockService.php
@@ -158,12 +158,12 @@ abstract class AbstractCategoriesBlockService extends AbstractClassificationBloc
     }
 
     /**
-     * @param CategoryInterface|int  $id
-     * @param CategoryInterface|null $default
+     * @param CategoryInterface|int $id
+     * @param mixed                 $default
      *
      * @return CategoryInterface
      */
-    final protected function getCategory($id, CategoryInterface $default = null)
+    final protected function getCategory($id, $default = null)
     {
         if ($id instanceof CategoryInterface) {
             return $id;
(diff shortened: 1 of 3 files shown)


greg0ire:
> The using the default value for category / collection / tag in the coresponding Abstract*BlockService an exception occurred:

I don't understand this sentence and I don't understand why there is a need to use `false` instead of `null` as a default value.

core23:
Updated the PR message. Sorry, got too many tasks in my head recently ☹️

greg0ire:
No problem. IIUC, [this](https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/Block/Service/AbstractTagsBlockService.php#L123) causes the problem, right? In that case, I think we should make sure the default value is not a boolean, but `null`. Also, we should configure the options resolver so that only instances of the right type are accepted (not sure if this is possible). But maybe we should merge this first, and kind of revert it in next major? I'm not sure…

jordisala1991:
Shouldn't we change default options `false` to `null` or that isn't possible without breaking BC?

core23:
> Shouldn't we change default options `false` to `null` or that isn't possible without breaking BC?

IMHO this is a BC break, that's why i changed the method.

greg0ire:
> But maybe we should merge this first, and kind of revert it in next major? I'm not sure…

So? We do that I guess?

greg0ire:
Thanks @core23

jordisala1991:
I think that should be the plan @greg0ire

PR #284, opened 2017-02-25
Fix bug with class_exists instead of interface_exists in CategorySele…

ZRazor:
Made a stupid mistake in my previous PR.
At the last moment I decided not to check on the existence of the class `SimpleChoiceList`, but on `ChoiceLoaderInterface` and forgot to replace `class_exists` with `interface_exists`
My fault.

### DIFF
--- a/Form/Type/CategorySelectorType.php
+++ b/Form/Type/CategorySelectorType.php
@@ -60,7 +60,7 @@ class CategorySelectorType extends AbstractType
     {
         $that = $this;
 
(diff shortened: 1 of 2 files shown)


greg0ire:
Thanks for fixing this quickly! Maybe we need some tests here :P

ZRazor:
@greg0ire, The test is there, but it contained this mistake too )
Thanks for quick review and merge!

greg0ire:
> @greg0ire, The test is there, but it contained this mistake too )

Oh that's right…

PR #285, opened 2017-03-01
Fix phpunit deprecated methods

ZRazor:
I am targetting this branch, because PR isn't break BC (should be still compatible with old phpunit versions)
## Changelog
```markdown
### Fixed
- Fixed usage of deprecated phpunit methods
```

PHPUnit 5.7.14
Before:
```
WARNINGS!
Tests: 88, Assertions: 256, Warnings: 72.
```
After:
```
OK (88 tests, 256 assertions)
```

### DIFF
--- a/Tests/Admin/AdminTest.php
+++ b/Tests/Admin/AdminTest.php
@@ -22,7 +22,7 @@ class AdminTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
-        $this->contextManager = $this->getMock('Sonata\ClassificationBundle\Model\ContextManagerInterface');
+        $this->contextManager = $this->getMockForAbstractClass('Sonata\ClassificationBundle\Model\ContextManagerInterface');
     }
 
     public function testAbstractAdminChildren()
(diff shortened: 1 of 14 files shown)


OskarStark:
Thank you @ZRazor 👍

Tests/Admin/AdminTest.php
```diff
@@ -22,7 +22,7 @@ class AdminTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
-        $this->contextManager = $this->getMock('Sonata\ClassificationBundle\Model\ContextManagerInterface');
+        $this->contextManager = $this->getMockForAbstractClass('Sonata\ClassificationBundle\Model\ContextManagerInterface');
```
core23: This is right for interfaces?

greg0ire: I wouldn't say it is right, but apparently, it works.

ZRazor: We cant use `createMock` because of BC with old PHPUnit versions.
This method is used for interfaces in other sonata bundles. Example from AdminBundle:
https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Tests/Admin/BaseFieldDescriptionTest.php#L83

jordisala1991: Not sure about this, IMO is better to do like this: https://github.com/sonata-project/SonataUserBundle/pull/840

Follows the styles of how we did on other bundles.

greg0ire: @jordisala1991 : this has been merged.

ZRazor: I can provide a new PR with `PHPUnit_Framework_TestCase` wrapper if needed.

core23: Would appreciate if you can create a new PR, so we have the same handling across all bundles.

ZRazor: @core23 I'm on it.

ZRazor: @core23 here it is #290

Issue #286, opened 2017-03-03
Datagrid filter conflicts with context parameter


ZRazor:
## Subject

Datagrid filter conflicts with context parameter.

## Steps to reproduce

For example we have 2 contexts: `default` and `context`.

When first entering the categories in the admin panel, the uri looks like:
`tree?hide_context=0`

We can switch contexts in categories context custom switcher
<img width="288" alt="2017-03-03 10 00 08" src="https://cloud.githubusercontent.com/assets/1560126/23532669/593a624c-fff8-11e6-8c2b-0fc5fa972b62.png">

If we switch to `default` context, the uri will look like this:
`tree?context=default&hide_context=0`

And the datagrid filter will appear with the selected context:
<img width="505" alt="2017-03-03 10 21 07" src="https://cloud.githubusercontent.com/assets/1560126/23533370/1d7a48ee-fffd-11e6-8531-c2dd93158c75.png">

If we select `context` context in datagrid filter and press "Filter", we will have uri:
`
list?filter[context][type]=&filter[context][value]=context&filter[name][type]=&filter[name][value]=&filter[enabled][type]=&filter[enabled][value]=&filter[_page]=1&filter[_sort_by]=id&filter[_sort_order]=ASC&filter[_per_page]=32&context=default&hide_context=0
`

Now we have `filter[context][value]=context` from the selected filter and `context=default`, which priority is higher.

## Expected results

We are in a tree mode with selected `context` context.

## Actual results

We are in a list mode with selected `default` context. Does not matter what is in the uri param `filter[context][value]`, filter will be set to `default` context, which is defined in `context=default` parameter. 
Now we can try to change context in datagrid filter, but it would be locked on `default` because of 
`context=default` parameter. 

If we remove `context=default` from uri, datagrid filter will work fine.

## Checked on
- last 3.x version and datagrid bundle 2.2.1
- 3.2.0 version and datagrid bundle 2.2.0

ZRazor:
Also in tree mode with context parameter (when datagrid filter appears)
`context=default&hide_context=0` we have a form (sonata_type_filter_default) error  in `context` field.
I take a look at it: 
in tree mode normalized format of this field looks like:
`[value => Object(Application\Sonata\ClassificationBundle\Entity\Context)]`
in list mode (where are no errors):
`[type => null, value => Object(Application\Sonata\ClassificationBundle\Entity\Context)]`

Conclusion:
`type` is missing.

[...]

PR #292, opened 2017-03-12
Fix missing type field in filters form in categories tree mode

ZRazor:
## Changelog

```markdown
### Fixed
- Fixed missing type field in filters form in categories tree mode
```

## Subject

I described this issue in my second comment in #286 
In tree mode `context` parameter should be set as it's set in list mode:
https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/Controller/CategoryAdminController.php#L45

### DIFF
--- a/Controller/CategoryAdminController.php
+++ b/Controller/CategoryAdminController.php
@@ -12,7 +12,6 @@
 namespace Sonata\ClassificationBundle\Controller;
 
 use Sonata\AdminBundle\Controller\CRUDController as Controller;
-use Sonata\AdminBundle\Form\Type\Filter\ChoiceType;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -99,7 +98,7 @@ class CategoryAdminController extends Controller
         $datagrid = $this->admin->getDatagrid();
 
         if ($this->admin->getPersistentParameter('context')) {
-            $datagrid->setValue('context', ChoiceType::TYPE_EQUAL, $this->admin->getPersistentParameter('context'));
+            $datagrid->setValue('context', null, $this->admin->getPersistentParameter('context'));
         }
 
         $formView = $datagrid->getForm()->createView();


greg0ire:
Please read and apply : https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/CONTRIBUTING.md#the-commit-message

People should be able to understand what is the bug and how your change fixes it just by reading the commit message.

ZRazor:
greg0ire, I changed commit message - it should be more specific now

Controller/CategoryAdminController.php
```diff
@@ -99,7 +98,7 @@ public function treeAction(Request $request)
         $datagrid = $this->admin->getDatagrid();
 
         if ($this->admin->getPersistentParameter('context')) {
-            $datagrid->setValue('context', ChoiceType::TYPE_EQUAL, $this->admin->getPersistentParameter('context'));
+            $datagrid->setValue('context', null, $this->admin->getPersistentParameter('context'));
```
greg0ire: I still don't understand… you say you are fixing a missing type, and the diff shows you are in fact removing it. What's the deal ? 😕

greg0ire: I guess the side effect of this is to let the TypeGuesser do its job, but how exactly does that help here?

ZRazor: I will debug and check datagrid-bundle code and say more specifically, what causes this problem.

If we pass `ChoiceType::TYPE_EQUAL` we will have the folowing situation:


When we pass `null` there we will have


And a few lines above - in list mode we already have such line:
https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/Controller/CategoryAdminController.php#L45

greg0ire: Ok so in the first screenshot you indeed have a form error you no longer have after… what does it say?

greg0ire: Maybe you can click the "+" sign to find out? Or dump `form_errors`?

ZRazor: I already checked - `type` is invalid

```
This value is not valid.	type	
Symfony\Component\Validator\ConstraintViolation
Object(Symfony\Component\Form\Form).children[context].children[type] = 3
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
Unable to reverse value for property path "[type]": The choice "3" does not exist or is not unique
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
The choice "3" does not exist or is not unique
```

greg0ire: I think [this is where the field is defined](https://github.com/ZRazor/SonataClassificationBundle/blob/3.x/Admin/ContextAwareAdmin.php#L106)

ZRazor: I found:
It is `Sonata\CoreBundle\Form\Type\EqualType` and we should pass there either `EqualType::TYPE_IS_EQUAL` or `EqualType::TYPE_IS_NOT_EQUAL`

ZRazor:
In `Sonata\AdminBundle\Form\Type\Filter\ChoiceType`:
```
    const TYPE_CONTAINS = 1;
    const TYPE_NOT_CONTAINS = 2;
    const TYPE_EQUAL = 3;
```
In `Sonata\CoreBundle\Form\Type\EqualType`:
```
    const TYPE_IS_EQUAL = 1;
    const TYPE_IS_NOT_EQUAL = 2;
```
I checked - this consts have never changed.
So it is incorrect to use `ChoiceType` instead of `EqualType`.

Controller/CategoryAdminController.php
greg0ire: Are you 100% sure? If yes, please pass `EqualType::TYPE_IS_EQUAL`, because it will help understand things faster if there ever is another bug in this area of the code.

ZRazor: Yes, I passed there `EqualType::TYPE_IS_EQUAL` and `EqualType::TYPE_IS_NOT_EQUAL` - no errors, everything works. Also if i pass there `ChoiceType::TYPE_CONTAINS` - there will be no errors too, because it it the same as to pass there `1`.  But is is incorrect to use consts from other types, even if they are equal.

greg0ire: Yeah, that's why using integers for consts is not great IMO. In my projects, I do `MY_CONST = 'my_const'`, way more clear / less collisions

greg0ire:
@sonata-project/contributors please review

Controller/CategoryAdminController.php
jordisala1991: So, we need to use EqualType::TYPE_IS_EQUAL right?

jordisala1991: Oh, I get it, we pass null because this is the default

jordisala1991:
Thanks @ZRazor !

Controller/CategoryAdminController.php
ZRazor: Documentation ( https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Resources/doc/reference/action_list.rst ) says that it should be `EqualType::TYPE_IS_EQUAL` but it is never passed in AdminBundle or MediaBundle, `null` is passed everywhere as default value.

PR #295, opened 2017-03-13
Fix conflict of datagrid context value and persistent parameter

### LINKED ISSUE #286: Datagrid filter conflicts with context parameter

ZRazor:
I am targeting this branch, because it is BC.

Closes #286

## Changelog

```markdown
### Fixed
- Fixed conflict of datagrid `context` value and persistent `context` parameter
```

### DIFF
--- a/Controller/CategoryAdminController.php
+++ b/Controller/CategoryAdminController.php
@@ -40,8 +40,12 @@ class CategoryAdminController extends Controller
         }
 
         $datagrid = $this->admin->getDatagrid();
+        $datagridValues = $datagrid->getValues();
 
-        if ($this->admin->getPersistentParameter('context')) {
+        $datagridContextIsSet = isset($datagridValues['context']['value']) && !empty($datagridValues['context']['value']);
+
+        //ignore `context` persistent parameter if datagrid `context` value is set
+        if ($this->admin->getPersistentParameter('context') && !$datagridContextIsSet) {
             $datagrid->setValue('context', null, $this->admin->getPersistentParameter('context'));
         }
 
--- a/Tests/Controller/CategoryAdminControllerTest.php
+++ b/Tests/Controller/CategoryAdminControllerTest.php
@@ -337,6 +337,14 @@ class CategoryAdminControllerTest extends PHPUnit_Framework_TestCase
             ->method('getForm')
             ->will($this->returnValue($form));
 
+        $datagrid->expects($this->once())
+            ->method('getValues')
+            ->will($this->returnValue(array(
+                'context' => array(
+                    'value' => $context ?: '',
+                ),
+            )));
+
         $this->admin->expects($this->any())
             ->method('getPersistentParameter')
             ->will($this->returnValue($context));


Controller/CategoryAdminController.php
```diff
@@ -40,8 +40,13 @@ public function listAction(Request $request = null)
         }
 
         $datagrid = $this->admin->getDatagrid();
+        $datagridValues = $datagrid->getValues();
 
-        if ($this->admin->getPersistentParameter('context')) {
+        $datagridContextIsNotSet = array_key_exists('context', $datagridValues) ?
```
greg0ire: Avoid negative boolean variables please, it hurts the brain : `$datagridContextIsSet = …`

ZRazor: Oh, I did it that way at first, but then changed to avoid `!empty` and `&& ! datagridContextIsSet`

Controller/CategoryAdminController.php
```diff
@@ -40,8 +40,13 @@ public function listAction(Request $request = null)
         }
 
         $datagrid = $this->admin->getDatagrid();
+        $datagridValues = $datagrid->getValues();
 
-        if ($this->admin->getPersistentParameter('context')) {
+        $datagridContextIsSet = array_key_exists('context', $datagridValues) ?
+            !empty($datagridValues['context']['value']) : false;
```
greg0ire: Maybe use `isset()` here then. BTW, why not simply use `isset()` directly, without the `array_key_exists` check?

ZRazor: We need to handle this situation:
```
'context' => array(
                'value' => ''
            )
```
`isset` will return true, which is not what we want

greg0ire: Then maybe `isset($datagridValues['context']['value']) && !empty($datagridValues['context']['value'])` ?

ZRazor: Yes, I agree this looks a bit better than `array_key_exists`.
I'll change it.

Tests/Controller/CategoryAdminControllerTest.php
```diff
@@ -337,6 +337,14 @@ public function testListAction($context)
             ->method('getForm')
             ->will($this->returnValue($form));
 
+        $datagrid->expects($this->once())
+            ->method('getValues')
+            ->will($this->returnValue(array(
+                'context' => array(
+                    'value' => $context ? $context : '',
```
greg0ire: I think you are allowed to use `$context ?: '',` here

ZRazor: Yes, just copied it from a few lines above

PR #299, opened 2017-03-28
Fix CategoryManager methods params types

ZRazor:
I am targeting this branch, because it's a bug fix.

## Changelog

```markdown
### Fixed
- Fixed `CategoryManager::getRootCategoriesForContext` to accept string codes of contexts
```

## Subject

Methods in `CategoryManager`, which have inside this line `$context = $this->getContext($context);` can accept not just instances of `ContextInterface`, but also string codes of contexts. For example:
`->categoryManager->getRootCategoriesForContext('product_catalog');`


Entity/CategoryManager.php
```diff
@@ -143,11 +143,11 @@ public function getRootCategory($context = null)
     /**
      * NEXT_MAJOR: add this method to the interface.
      *
-     * @param ContextInterface $context
+     * @param ContextInterface|string|null $context
      *
      * @return CategoryInterface[]
      */
-    public function getRootCategoriesForContext(ContextInterface $context = null)
+    public function getRootCategoriesForContext($context = null)
```
greg0ire: This is a BC-break for child classes that redefine this method : https://3v4l.org/1paGC

ZRazor: You wanted to make these methods final, but we didn't do it, because they were stubbed in tests.
I really don't think, that anyone has overridden this method after the latest release. But if it's critical, I can change it to master branch.

greg0ire: > because they were stubbed in tests...

How hard would it be to change from stubs to real objects? Or to introduce an interface you could mock?

ZRazor: I checked through:
 We stub only one method `CategoryManager::getRootCategoriesSplitByContexts` in `CategoryAdminControllerTest`, and I'm sure that we can change mocking of `CategoryManager` it to a real object (How it's done in `CategoryManagerTest`)

greg0ire: Then I think it's safe to turn a blind eye to the BC-break and make the method final. This will help us reason safely in the future (for instance, when adding return type hinting for php 7)

OskarStark:
@ZRazor could you finish this PR?

Entity/CategoryManager.php
```diff
@@ -127,7 +127,7 @@ public function getRootCategoryWithChildren(CategoryInterface $category)
     /**
      * @todo NEXT_MAJOR: remove this method.
      *
-     * @param ContextInterface $context
+     * @param ContextInterface|string|null $context
```
core23: Instead of supporting different types (objects and strings), could you please pick one and deprecate the other?

core23:
This will be deprecated with #464

PR #304, opened 2017-04-20
#303: Changed deprecated twig test from 'sameas' to 'same as'.

### LINKED ISSUE #303: Deprecated twig test 'sameas' in SonataClassificationBundle:Block:base_block_categories.html.twig
Twig template is using sameas test what is deprecated since Twig version 1.21.

```
Twig Test "sameas" is deprecated since version 1.21. Use "same as" instead in SonataClassificationBundle:Block:base_block_categories.html.twig at line 38.
```

gbalcewicz:
I am targeting this branch, because it's BC.

Closes #303

## Changelog
```markdown
### Added
- Added explicit `twig/twig` dependency
### Fixed
- use `same as` instead of deprecated `sameas` in twig template
```

### DIFF
--- a/Resources/views/Block/base_block_categories.html.twig
+++ b/Resources/views/Block/base_block_categories.html.twig
@@ -35,8 +35,8 @@ file that was distributed with this source code.
 
 {% macro attributes(attributes) %}
     {% for name, value in attributes %}
-        {%- if value is not none and value is not sameas(false) -%}
-            {{- ' %s="%s"'|format(name, value is sameas(true) ? name|e : value|e)|raw -}}
+        {%- if value is not none and value is not same as(false) -%}
+            {{- ' %s="%s"'|format(name, value is same as(true) ? name|e : value|e)|raw -}}
         {%- endif -%}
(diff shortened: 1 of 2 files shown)


greg0ire:
Should we require Twig >= 1.14.2 , @sonata-project/contributors ? Right now there is no twig requirement, not sure why… The "same as" alias was introduced in that version, that's why I'm asking.

gbalcewicz:
There is requirement for sonata-admin in 3.1 what requires  twig ^1.23. So I think it's covered there.

greg0ire:
> So I think it's covered there.

It's implicitely covered, so we're accidentally fine indeed, but it would be best if we were explicit IMO. You know, in case we suddenly need to bump it to 1.24 or whatever.

jordisala1991:
Let's add twig requirement if we use twig.

gbalcewicz:
I have added twig requirement.

greg0ire:
Thanks @gbalcewicz ! Nice first contribution!

PR #324, opened 2017-08-29
Remove exceptions from phpunit whitelist

jordisala1991:
I am targeting this branch, because this is pedantic.

## Subject

I think coverage is not calculated correctly, lets see

### DIFF
--- a/Controller/CategoryAdminController.php
+++ b/Controller/CategoryAdminController.php
@@ -75,7 +75,7 @@ class CategoryAdminController extends Controller
             $currentContext = $this->get('sonata.classification.manager.context')->find($context);
         }
 
-        // all root categories
+        // all root categories.
         $rootCategoriesSplitByContexts = $categoryManager->getRootCategoriesSplitByContexts(false);
 
         // root categories inside the current context
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -13,7 +13,7 @@
 >
     <testsuites>
         <testsuite name="ClassificationBundle Test Suite">
-            <directory>./Tests/</directory>
+            <directory>./Tests</directory>
         </testsuite>
     </testsuites>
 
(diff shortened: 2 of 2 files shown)


phpunit.xml.dist
```diff
@@ -22,14 +22,16 @@
...
-                <!-- Ignore ODM driver -->
-                <directory>./Document/</directory>
+                <directory>./coverage/</directory>
             </exclude>
         </whitelist>
     </filter>
 
+    <php>
+        <ini name="precision" value="8"/>
+    </php>
```
greg0ire: This does not match your commit message.

Controller/CategoryAdminController.php
```diff
@@ -75,7 +75,7 @@ public function treeAction(Request $request)
             $currentContext = $this->get('sonata.classification.manager.context')->find($context);
         }
 
-        // all root categories
+        // all root categories.
```
greg0ire: I suppose this comes from StyleCI?

phpunit.xml.dist
jordisala1991: Not sure what it does, I am trying to make all phpunit configs look the same, copied from SonataAdminBundle

Controller/CategoryAdminController.php
jordisala1991: This is needed because otherwise the build will not run, phpunit.xml is not part of the relevant files

phpunit.xml.dist
```diff
@@ -22,14 +22,16 @@
             <directory>./</directory>
             <exclude>
                 <directory>./Tests/</directory>
+                <directory>./DataFixtures/</directory>
                 <directory>./Resources/</directory>
-                <directory>./DependencyInjection/</directory>
                 <directory>./vendor/</directory>
-
-                <!-- Ignore ODM driver -->
```
OskarStark: hmm strange comment, can you find out why this was explicitely excluded @jordisala1991?

jordisala1991: Not sure why it was removed, seems its been here since the begining of this repository.

For me its like excluding entities from the coverage

OskarStark: 👍

OskarStark:
Thank you @jordisala1991 👍

PR #363, opened 2017-12-08
Added missing getId method to Context model

core23:
I am targeting this branch, because this is a patch.

```markdown
### Added
- Added `Context::getId` method
```

## Subject

There is already a `setId` method, but the getter was missing.

### DIFF
--- a/src/Model/Context.php
+++ b/src/Model/Context.php
@@ -122,4 +122,12 @@ abstract class Context implements ContextInterface
     {
         $this->id = $id;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getId()
+    {
+        return $this->id;
+    }
 }


greg0ire:
Not sure it was overlooked, maybe it is never needed?

core23:
The method exists in the interface... https://github.com/sonata-project/SonataClassificationBundle/blob/eaf69f12198a56a8edcf18b1e8557ea208de4364/src/Model/ContextInterface.php#L58

jordisala1991:
I think this happens on all sonata entities. And it is an abstract class, so when you extend it you normally add the method. Not saying not to this PR but reconsider if it is really needed, or why it wasnt there

core23:
Yes, it's an abstract class, but all methods (except this one) is implemented, even the setter method.

core23:
Please review @jordisala1991

jordisala1991:
Already did, I am not sure why do we need this method. It was approved by @soullivaneuh and @greg0ire , so they can merge it if they want.

greg0ire:
I think this is a sensible implementation.

PR #365, opened 2017-12-08
Fix context id type hint and validation

gremo:
I am targeting this branch, because this is a minor fix.

## Changelog

```markdown
### Added
- NotBlank constraint: context id property is now required
```

## Subject
Context id should be a string, type hint is wrong in both `Context ` and `ContextInterface`. Property id is missing the `NotBlank` validation constraint.

Without the constraint when you create a new context and forget to set the id, an exception is thrown. #

### DIFF
--- a/src/Model/ContextInterface.php
+++ b/src/Model/ContextInterface.php
@@ -44,16 +44,16 @@ interface ContextInterface
     public function getEnabled();
 
     /**
-     * Set slug.
+     * Set id.
      *
-     * @param int $id
+     * @param string $id
      */
     public function setId($id);
 
     /**
-     * Get slug.
+     * Get id.
      *
-     * @return int $slug
+     * @return string $id
      */
     public function getId();
 
--- a/src/Model/Context.php
+++ b/src/Model/Context.php
@@ -14,7 +14,7 @@ namespace Sonata\ClassificationBundle\Model;
 abstract class Context implements ContextInterface
 {
     /**
-     * @var int
+     * @var string
      */
     protected $id;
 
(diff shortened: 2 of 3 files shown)


src/Model/ContextInterface.php
```diff
@@ -44,16 +44,16 @@ public function setEnabled($enabled);
     public function getEnabled();
 
     /**
-     * Set slug.
+     * Set id.
```
greg0ire: Please remove that instead.

src/Model/ContextInterface.php
```diff
@@ -44,16 +44,16 @@ public function setEnabled($enabled);
...
      *
-     * @param int $id
+     * @param string $id
      */
     public function setId($id);
 
     /**
-     * Get slug.
+     * Get id.
```
greg0ire: And that too. We learn nothing here.

src/Model/ContextInterface.php
gremo: What should I remove? The entire phpdoc block? Why?

greg0ire: Just the description, because it is tautological

greg0ire: See https://github.com/sonata-project/dev-kit/issues/345 and CONTRIBUTING.md

src/Model/ContextInterface.php
gremo: @greg0ire agree, but it should be done on the entire project. What about another PR in the future, leaving it as-is for now?

src/Model/ContextInterface.php
OskarStark: I think @gremo just wanted to fix the copy paste mistake!

src/Model/ContextInterface.php
greg0ire: As you wish

PR #367, opened 2017-12-08
Make the context foreign key not nullable in Tag entity

### LINKED ISSUE #366: Context is required when creating a tag, but this isn't enforced at database level
As per validation rules ([validation.xml](https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/src/Resources/config/validation.xml#L22)) a tag cannot be created without a context. This is not enforced at database level (foreign key can be null).

Should we make the column `NOT NULL` and add a `ON DELETE CASCADE` behaviour on it (currently is `RESTRICT` see https://github.com/sonata-project/SonataClassificationBundle/issues/361)?

gremo:
I am targeting this branch, because this is an enhancement.

Closes #366

## Changelog

```markdown
### Changed
- Column `context` for entity Tag is now not nullable
- When you delete a context all tags that belongs to that context are removed
```

## Subject
A tag cannot be created without a context. So context column should be `NOT NULL` and delete behaviour should be `ON DELETE CASCADE`.

### DIFF
--- a/src/DependencyInjection/SonataClassificationExtension.php
+++ b/src/DependencyInjection/SonataClassificationExtension.php
@@ -175,6 +175,8 @@ class SonataClassificationExtension extends Extension
                 [
                     'name' => 'context',
                     'referencedColumnName' => 'id',
+                    'nullable' => false,
+                    'onDelete' => 'CASCADE',
                 ],
             ],
             'orphanRemoval' => false,


core23:
Why did you only add this to the `tag` and not to `category` and `collection`?

gremo:
@core23 you are missing the whole point, I'm going to create just one issue/PR, they are all related (see my comment here https://github.com/sonata-project/SonataClassificationBundle/issues/366#issuecomment-350451674).

gremo:
Closed in favour of https://github.com/sonata-project/SonataClassificationBundle/issues/386

PR #368, opened 2017-12-08
Make the context foreign key not nullable in Collection entity

gremo:
I am targeting this branch, because this is an enhancement.

Similar of #367.

## Changelog

```markdown
### Changed
- Column `context` for entity Collection is now not nullable
- When you delete a context all collections that belongs to that context are removed
```

## Subject
A collection cannot be created without a context. So context column should be `NOT NULL` and delete behaviour should be `ON DELETE CASCADE`.

### DIFF
--- a/src/DependencyInjection/SonataClassificationExtension.php
+++ b/src/DependencyInjection/SonataClassificationExtension.php
@@ -194,6 +194,8 @@ class SonataClassificationExtension extends Extension
                 [
                     'name' => 'context',
                     'referencedColumnName' => 'id',
+                    'nullable' => false,
+                    'onDelete' => 'CASCADE',
                 ],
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/SonataClassificationExtension.php
```diff
@@ -194,6 +194,8 @@ public function registerDoctrineMapping(array $config)
                 [
                     'name' => 'context',
                     'referencedColumnName' => 'id',
+                    'nullable' => false,
```
core23: What about the other parts (tag, category)?

gremo:
Closed in favour of https://github.com/sonata-project/SonataClassificationBundle/issues/386

PR #375, opened 2017-12-21
Allow Slugify 3.0 and higher

### LINKED ISSUE #374: Slugify 3.0
## Subject

Allow cocur/slugify 3.0 since the bundle supports up to 2.5 only

## Steps to reproduce

Try to install cocur/slugify 3.0 when SonataClassificationBundle is installed

## Expected results

It installs

## Actual results

It doesn't install cause ^2.0 is maximum in composer.json

KDederichs:
I am targeting this branch, because it causes no BC break.

Closes #374

## Changelog

Make the bundle compatible with Slugify ^3.0

```markdown

### Changed
Slugify dependency 
```
## Subject

Add ^3.0 as compatible Slugify version so people who use SonataClassificationBundle can also use Slugify in version 3.0 and higher

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": "^5.6 || ^7.0",
-        "cocur/slugify": "^1.4 || ^2.0",
+        "cocur/slugify": "^1.4 || ^2.0 || ^3.0",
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @KDederichs !

PR #379, opened 2018-01-16
Sets the service sonata.classification.manager.category service to public

BOAWEB:
I kept getting the message 'sonata.classification.manager.category' service or alias has been removed or inlined when the container was compiled. This solves it.

I am targeting this branch, because this branch should be SF4 compatible.

## Changelog

```markdown

### Fixed
- Make `sonata.classification.manager.category` public

```
## Subject

In SF4 I kept getting the message: 'The "sonata.classification.manager.category" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.'

So, I listened to the message and made it public.

### DIFF
--- a/src/Resources/config/orm.xml
+++ b/src/Resources/config/orm.xml
@@ -7,7 +7,7 @@
         <parameter key="sonata.classification.manager.context.class">Sonata\ClassificationBundle\Entity\ContextManager</parameter>
     </parameters>
     <services>
-        <service id="sonata.classification.manager.category" class="%sonata.classification.manager.category.class%">
+        <service id="sonata.classification.manager.category" class="%sonata.classification.manager.category.class%" public="true">
             <argument>%sonata.classification.manager.category.entity%</argument>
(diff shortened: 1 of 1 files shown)


OskarStark:
Thank you @BOAWEB 👍

BOAWEB:
Can anybody restart the build, please? It failed because packagist.org was down.

greg0ire:
What build? this is merged...

BOAWEB:
Yes, this is merged, but the Travis CI build failed. 
image
Build #927.7 to be precise. As you can see in the logs, packagist was the problem. It's solved now.
But because the build is in error, no new version is created to use in composer.

greg0ire:
Ah right. But there is no relation between those builds and releases. Releases are not automatic.

greg0ire:
Restarted said job

BOAWEB:
Oh wow, sorry for the misunderstanding then. But thanks for restarting :)

PR #387, opened 2018-02-04
Fix context associations

### LINKED ISSUE #386: Associations inconsistency for tag, collection and category towards context
I'm creating this issue as an explanation about my upcoming PR.

**Fact 1**: as per validation rules, a **tag cannot be created without a context**.

https://github.com/sonata-project/SonataClassificationBundle/blob/d189426c05ed3fa5def926cd0f24c34585439cfa/src/Resources/config/validation.xml#L20-L22

However:
1. This is not enforced at database level because the foreign key _can_ be null.
2. The current foreign key behavior is `RESTRICT`, that means that a context can't be deleted if a tag is associated to that context. When you try to delete a context already associated to a tag an exception is thrown (at database level).

**Proposal 1**: make the `context` column `NOT NULL` and add the behavior `ON DELETE CASCADE`, because no one should have a tag or collection without a context (at least, using the admin).

**Fact 2**: same as fact 1, but for **collection**.
**Proposal 2**: same as proposal 1.

**Fact 3**: when you create a  category, **a context is always associated to that category** (but pay attention to the point 2).

https://github.com/sonata-project/SonataClassificationBundle/blob/d189426c05ed3fa5def926cd0f24c34585439cfa/src/Admin/ContextAwareAdmin.php#L46-L51

However:
1. The current foreign key behavior is `RESTRICT`, that means that a context can't be deleted if a category is associated to that context.
2. If there are no categories at all, the context parameter is absent and the context isn't created nor associated with the category. Thus, after adding a new root category (which is missing the context), an exception is thrown when browsing category tree, because of this check:

https://github.com/sonata-project/SonataClassificationBundle/blob/d189426c05ed3fa5def926cd0f24c34585439cfa/src/Entity/CategoryManager.php#L175-L177

**Proposal 3**:  make the `context` column `NOT NULL` and add the behaviour `ON DELETE CASCADE`; always create a context in `ContextAwareAdmin::getNewInstance()` even if the context parameter is missing.

**Bonus point**: the `sonata:classification:fix-context` command can be deleted because we create context and always assign a context to tags, collections and categories using the admin.

gremo:
I am targeting this branch, because this is a bugfix and an improvement.

Closes #386

## Changelog

```markdown
### Added
- Creation of new default context when context parameter is missing

### Changed
- Foreign keys towards context are now not nullable

### Fixed
- Category parent selector shown when not necessary 
```

### DIFF
--- a/tests/Model/CategoryTest.php
+++ b/tests/Model/CategoryTest.php
@@ -67,8 +67,12 @@ class CategoryTest extends TestCase
 
     public function testParent()
     {
+        /** @var ContextInterface $context */
+        $context = $this->createMock(ContextInterface::class);
+
         /** @var Category $parent */
         $parent = $this->getMockForAbstractClass(Category::class);
+        $parent->setContext($context);
 
         /** @var Category $category */
         $category = $this->getMockForAbstractClass(Category::class);


src/Admin/CategoryAdmin.php
```diff
@@ -63,6 +63,10 @@ public function getFormBuilder()
      */
     protected function configureFormFields(FormMapper $formMapper)
     {
+        $categoryManager = $this->getConfigurationPool()
+            ->getContainer()
+            ->get('sonata.classification.manager.category');
```
greg0ire: This should be obtained with constructor injection.

gremo:
Any update on this?

src/Admin/CategoryAdmin.php
greg0ire: @gremo what about this?

gremo: @greg0ire no problem I can change it, I was talking about the entire PR... what do you think?

greg0ire: Uuuuh it looks good I guess

src/Admin/CategoryAdmin.php
```diff
@@ -30,6 +32,23 @@ class CategoryAdmin extends ContextAwareAdmin
...
+     * @var CategoryManagerInterface
+     */
+    protected $categoryManager;
+
+    public function __construct(
+        $code,
+        $class,
+        $baseControllerName,
+        ContextManagerInterface $contextManager,
+        CategoryManagerInterface $categoryManager
```
core23: This is a BC break, if this field is required

jordisala1991: Maybe make it nullable and set it with container if null. With a next major message to remove that code?

gremo: @jordisala1991 see my update, you mean like this?

jordisala1991: Yes, exactly. I am not an expert of this bundle, that's why I will let others approve this, but this should make this BC

src/Admin/CategoryAdmin.php
```diff
@@ -30,6 +32,23 @@ class CategoryAdmin extends ContextAwareAdmin
         'cascade_validation' => true,
     ];
 
+    /**
+     * @var CategoryManagerInterface
+     */
+    protected $categoryManager;
+
+    public function __construct(
```
core23: Please add PHPDoc where we don't have an explicit type in the method/constructor signature

src/Admin/ContextAwareAdmin.php
```diff
@@ -43,22 +43,24 @@ public function getNewInstance()
...
+        $context = $this->contextManager->find($contextId);
+        if (!$context) {
+            /** @var ContextInterface $context */
+            $context = $this->contextManager->create();
+            $context->setId($contextId);
+            $context->setName(ucfirst($contextId));
+            $context->setEnabled(true);
 
-            $instance->setContext($context);
+            $this->contextManager->save($context);
```
core23: Will this create a new context object in the database everytime you open create form? If yes could you move this to the `prePersist` section, so we only create a context object if we really have an entity.

src/Entity/CategoryManager.php
```diff
@@ -105,9 +105,6 @@ public function getSubCategoriesPager($categoryId, $page = 1, $limit = 25, $crit
      */
     public function getRootCategoryWithChildren(CategoryInterface $category)
     {
-        if (null === $category->getContext()) {
```
core23: Why did you remove this? This can still happen if you've create a category outside the admin panel

gremo: @core23 your right, let me push ah update...

gremo: Forget what I said, context can be null (at database level) for category. So It shouldn't happen, why you say it's possible?

src/Admin/ContextAwareAdmin.php
gremo: Yes, this is a good idea. I'll try.

gremo: Ok @core23 this is too risky. Trying to comment $this->contextManager->save($context), lot of errors creating (for example, a tag). PrePersist isn't a choice because the context validation happens before.

src/Entity/CategoryManager.php
gremo: @core23 ping

PR #396, opened 2018-04-21
Fix tests

kunicmarko20:
## Subject

Template Registry broke some tests.


tests/Controller/CategoryAdminControllerTest.php
```diff
@@ -277,6 +282,8 @@ protected function setUp()
                         return $categoryManager;
                     case 'sonata.classification.manager.context':
                         return $contextManager;
+                    case 'admin_code.template_registry':
+                        return $this->createMock(TemplateRegistryInterface::class);
```
greg0ire: Have you tried `new TemplateRegistry` instead?

kunicmarko20: I just saw everything is mocked and went with it, changed and works.

PR #399, opened 2018-05-23
Stop using Sonata\AdminBundle\Controller\CRUDController::render and use renderWithExtraParams

enekochan:
I am targeting this branch, because is Deprecation and BC.

## Changelog

```markdown
### Changed
- Make `Sonata\ClassificationBundle\Controller\CategoryAdminController` use `renderWithExtraParams`
```

## Subject
Stop using Sonata\AdminBundle\Controller\CRUDController::render and use renderWithExtraParams

### DIFF
--- a/src/Controller/CategoryAdminController.php
+++ b/src/Controller/CategoryAdminController.php
@@ -59,7 +59,7 @@ class CategoryAdminController extends Controller
         // set the theme for the current Admin Form
         $this->setFormTheme($formView, $this->admin->getFilterTheme());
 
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @enekochan !

greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

PR #402, opened 2018-06-04
Make services public

covex-nn:
I am targeting this branch, because this is a bugfix.

## Changelog

```markdown
### Fixed
- Make entity managers services public
```


greg0ire:
Thanks @covex-nn !

PR #404, opened 2018-06-13
Disallow null value on foreign keys for associations with Context

covex-nn:
I am targeting this branch, because existing application could be broken after applying proposed changes.

## Changelog

```markdown
### Changed
- Disallow null value on foreign keys for associations with Context
```

## Subject

With this PR it won't be possible to create a `Category`, `Tag` and `Collection` with null `$context`. It was possible when ClassificationBundle is used without MediaBundle. After such Category created, there is a error Context cannot be null on a tree view. I created a recipe for ClassificationBundle, see https://github.com/symfony/recipes-contrib/pull/389, where i also added a constraint NotNull to `$context` property.

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -26,3 +26,7 @@ If you have implemented a custom model, you must adapt the signature of the foll
  * `getRootCategoriesForContext`
  * `getAllRootCategories`
  * `getRootCategoriesSplitByContexts`
+
+## Doctrine schema update
+Disallowed null value on foreign keys for associations with Context for
+Category, Tag and Collection entities.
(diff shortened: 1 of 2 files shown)


UPGRADE-4.0.md
```diff
@@ -26,3 +26,7 @@ If you have implemented a custom model, you must adapt the signature of the foll
  * `getRootCategoriesForContext`
  * `getAllRootCategories`
  * `getRootCategoriesSplitByContexts`
+
+## Doctrine schema update
+Disallowed null value on foreign keys for associations with Context for
+Category, Tag and Collection entities.
```
greg0ire: Please do not edit this file. Instead, add an upgrade note in the first message of your PR.

kunicmarko20:
Thanks @covex-nn

PR #463, opened 2019-09-08
Use more specific PHPDoc

core23:
## Subject

I am targeting this branch, because this is a patch.

## Changelog

```markdown
### Added
- Add more `@method` annotation to propagate new methods of `CategoryManagerInterface`

### Removed
- Remove superfluous PHPDoc 

### Fixed
- Match PHPDoc with doctrine model 
```

### DIFF
--- a/src/Entity/CategoryManager.php
+++ b/src/Entity/CategoryManager.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sonata\ClassificationBundle\Entity;
 
 use Doctrine\Common\Persistence\ManagerRegistry;
-use Sonata\AdminBundle\Datagrid\PagerInterface;
 use Sonata\ClassificationBundle\Model\CategoryInterface;
 use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
 use Sonata\ClassificationBundle\Model\ContextInterface;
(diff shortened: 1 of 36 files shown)


src/Entity/CategoryManager.php
```diff
@@ -155,13 +120,6 @@ public function getRootCategoriesForContext(ContextInterface $context = null)
         return $this->categories[$context->getId()];
     }
 
-    /**
-     * @todo NEXT_MAJOR: remove this method.
```
OskarStark: Shouldn’t we keep this?

core23: IMHO we don't need this. We just need to port all `@method` methods

src/Entity/CategoryManager.php
```diff
@@ -48,12 +47,6 @@ public function __construct($class, ManagerRegistry $registry, ContextManagerInt
 
     /**
      * Returns a pager to iterate over the root category.
-     *
-     * @param int   $page
```
OskarStark: Why removing this? There is no typehint so this is the only place to get the type, isn’t it?

src/Entity/CategoryManager.php
OskarStark: For adding yes, but this is about removing a method

core23: Oh, you're right

src/Entity/CategoryManager.php
core23: The type hint was moved: https://github.com/sonata-project/SonataClassificationBundle/pull/463/commits/e5739d249bf334ddb3f49c6df79c804d12f2dd10#diff-4d9dd8b669cb2a016d76d86e7d6f0384R20

src/Entity/CategoryManager.php
OskarStark: There are others too IIRC

core23: Fixed

PR #466, opened 2019-09-11
Add strict types to CategoryManager

core23:
## Subject

I am targeting this branch, because this is a minor change.

## Changelog

```markdown
### Added
- Add strict types to CategoryManager

### Changed
- `CategoryManager::getRootCategory` will throw an exception if you pass invalid arguments
```

### DIFF
--- a/src/Entity/CategoryManager.php
+++ b/src/Entity/CategoryManager.php
@@ -89,7 +89,8 @@ class CategoryManager extends BaseEntityManager implements CategoryManagerInterf
         if (null !== $category->getParent()) {
             throw new \RuntimeException('Method can be called only for root categories');
         }
-        $context = $this->getContext($category->getContext());
+
+        $context = $category->getContext();
 
         $this->loadCategories($context);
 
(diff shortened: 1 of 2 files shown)


src/Entity/CategoryManager.php
```diff
@@ -105,13 +106,20 @@ public function getRootCategoryWithChildren(CategoryInterface $category)
...
     {
-        $context = $this->getContext($context);
+        if (null === $context || \is_string($context)) {
+            $context = $this->getContext($context);
+        } elseif (!$context instanceof ContextInterface) {
+            throw new \InvalidArgumentException(sprintf(
+                'Invalid parameter given: %s',
+                (string) $context
+            ));
+        }
```
greg0ire: Maybe avoid the repetition with
```
private function fromMixedTypeContext(?$context = null): ContextInterface
{
        if (null === $context || \is_string($context)) {
            return $this->getContext($context);
        }

        if ($context instanceof ContextInterface) {
            return $context;
        }

        throw new \InvalidArgumentException(sprintf(
            'Invalid parameter given: %s',
            (string) $context
        ));
}
```

greg0ire:
Thanks @core23 !

PR #471, opened 2019-10-12
Update CategoryManager.php

### LINKED ISSUE #470: Empty context causes infinite loop
#### Sonata packages
sonata-project/classification-bundle: 3.9.0

#### Symfony packages
n/a

#### PHP version
7.3

## Subject

Due to the bug within code, context is not being created causing infinite loop.

## Steps to reproduce
1. Issue \Sonata\ClassificationBundle\Entity\CategoryManager::getRootCategory with empty database

## Expected results
Context being created.

## Actual results
Infinite entity loop.

I've dug into the code and solution is trivial: `\Sonata\ClassificationBundle\Entity\contextObj:299`. The name of context is being overwritten by object which is also used as name. It produces weird results. I've created a pull request for fixing this issue.

er1z:
## Subject

getContext caused name/id to be self-recursiveI am targeting this branch, because this is backwards-compatible.

Closes #470, #472 

## Changelog

```markdown
### Fixed
- `getContext` method with non-existing one causing self-recursiveness
```

### DIFF
--- a/src/Entity/CategoryManager.php
+++ b/src/Entity/CategoryManager.php
@@ -293,18 +293,18 @@ class CategoryManager extends BaseEntityManager implements CategoryManagerInterf
             ));
         }
 
-        $context = $this->contextManager->find($context);
+        $contextObj = $this->contextManager->find($context);
 
-        if (!$context instanceof ContextInterface) {
-            $context = $this->contextManager->create();
+        if (!$contextObj instanceof ContextInterface) {
+            $contextObj = $this->contextManager->create();
 
-            $context->setId($context);
-            $context->setName($context);
-            $context->setEnabled(true);
+            $contextObj->setId($context);
+            $contextObj->setName($context);
+            $contextObj->setEnabled(true);
 
-            $this->contextManager->save($context);
+            $this->contextManager->save($contextObj);
         }
 
-        return $context;
+        return $contextObj;
     }
 }


greg0ire:
Hi! What is the difference between recursive and self-recursive?

greg0ire:
Also, can you provide the execution path that suffers from recursivity / a stack trace showing the bug?

src/Entity/CategoryManager.php
```diff
@@ -293,18 +293,18 @@ private function getContext($context = null): ContextInterface
...
-        $context = $this->contextManager->find($context);
+        $contextObj = $this->contextManager->find($context);
 
-        if (!$context instanceof ContextInterface) {
-            $context = $this->contextManager->create();
+        if (!$contextObj instanceof ContextInterface) {
+            $contextObj = $this->contextManager->create();
 
-            $context->setId($context);
-            $context->setName($context);
```
greg0ire: Oh right these to lines seem indeed problematic, good job!

er1z:
May be I misused the words. The clue is fact that if you have an empty database and desired context does not exist, `getContext` method tries to create a new entity. The problem is that a newly created object is assigned to `$context` argument. Then you try to assign an object to the name (which no longer exists) that expects a string.

greg0ire:
Yeah I was confused but after reading the 2 lines above I get it 👍

enemis:
As i understood the idea of usage context entity i would give the users possibility doesnt have in their databases 'default' context. I think, it would be better if in that place in codebase where @er1z had added fix, before creating default context the manager calls findAll method and returns first context. And only if the result from findAll() will be empty it should create default context.

PR #479, opened 2019-11-11
Dropping support for symfony < 3.4

franmomu:
## Subject

This is the same thing as https://github.com/sonata-project/SonataAdminBundle/pull/5733, dropping support of Symfony < 3.4 and >= 4, < 4.2

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Removed
- Support for Symfony < 3.4
- Support for Symfony >= 4, < 4.2
```

There are some changes that weren't marked as deprecated:

- [Removing `getName` from Form](https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.0.md#form)


tests/Controller/CategoryAdminControllerTest.php
```diff
@@ -189,50 +178,26 @@ protected function setUp()
...
-                    return 'csrf-token-123_'.$intention;
-                });
-
-            $this->csrfProvider->expects($this->any())
-                ->method('isCsrfTokenValid')
-                ->willReturnCallback(static function ($intention, $token) {
-                    if ($token === 'csrf-token-123_'.$intention) {
-                        return true;
-                    }
+        $this->csrfProvider = $this->getMockBuilder(
```
OskarStark: Can we use createMock() here?

src/Form/Type/CategorySelectorType.php
```diff
@@ -42,32 +39,10 @@ public function __construct(ManagerInterface $manager)
         $this->manager = $manager;
     }
 
-    /**
-     * NEXT_MAJOR: Remove method, when bumping requirements to SF 2.7+.
-     */
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
```
phansys: Removing a public method breaks BC.

phansys: We should deprecate these methods.

greg0ire:
Thanks @franmomu !

PR #482, opened 2019-11-17
Leverage `$this` binding at class scoped closures

phansys:
## Subject

Leverage `$this` binding at class scoped closures

I am targeting this branch, because this change respects BC.

[ref.](https://www.php.net/manual/en/functions.anonymous.php):
>As of PHP 5.4.0, when declared in the context of a class, the current class is automatically bound to it, making $this available inside of the function's scope. If this automatic binding of the current class is not wanted, then static anonymous functions may be used instead.

## Changelog

```markdown
### Changed
- Static closure to regular one in order to use `$this` from the class scope.
```


greg0ire:
Thanks @phansys !

PR #490, opened 2020-01-26
Fix media bundle decoupling

core23:
## Subject

This is just an other prepartion for the next major version (https://github.com/sonata-project/SonataClassificationBundle/pull/196)

## Changelog

```markdown
### Changed
- Loading media association based on defined class

### Fixed
- Fix media bundle decoupling
```


core23:
Changed the label to `minor` as this is a soft BC break, if you rely on the default value

PR #502, opened 2020-02-12
Remove SonataEasyExtendsBundle

franmomu:
## Subject

I replaced some docs, I'm not using the bundle so I'm not sure if is alright.

Refs https://github.com/sonata-project/SonataEasyExtendsBundle/issues/237

## Changelog

```markdown
### Removed
- SonataEasyExtendsBundle
```


core23:
I'm not sure if we should do this on a stable branch. Some default config values will change and you are about to remove functionality.

docs/reference/installation.rst
```diff
@@ -62,95 +59,192 @@ Add these bundles in the config mapping definition (or enable `auto_mapping`_):
...
 .. note::
 
     If you are not using Symfony Flex, this configuration should be added
     to ``app/config/config.yml``.
 
-Extending the Bundle
---------------------
-At this point, the bundle is functional, but not quite ready yet. You need to
-generate the correct entities for the media:
+And then create the corresponding entities, ``src/Entity/Tag``:
```
OskarStark:
```suggestion
And then create the corresponding entities, ``src/Entity/Tag``::
```

docs/reference/installation.rst
```diff
@@ -62,95 +59,192 @@ Add these bundles in the config mapping definition (or enable `auto_mapping`_):
...
 
-.. code-block:: bash
+.. code-block:: php
```
OskarStark:
```suggestion

```

src/DependencyInjection/SonataClassificationExtension.php
```diff
@@ -107,131 +108,145 @@ public function registerDoctrineMapping(array $config)
 
         $collector = DoctrineCollector::getInstance();
 
-        $collector->addAssociation($config['class']['category'], 'mapOneToMany', [
-            'fieldName' => 'children',
-            'targetEntity' => $config['class']['category'],
-            'cascade' => [
+        $categoryOnyToManyOptions = OptionsBuilder::create()
```
core23: When using the new doctrine extensions, you should check if the `SonataDoctrineBundle` bundle is loaded. Otherwise the defined associations are ignored, because you are doing a static call, not a symfony service call.

src/DependencyInjection/SonataClassificationExtension.php
```diff
@@ -52,7 +53,10 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('admin.xml');
         }
 
-        $this->registerDoctrineMapping($config);
+        if (isset($bundles['SonataDoctrineBundle'])) {
+            $this->registerDoctrineMapping($config);
```
core23: I would expect an exception, if the bundle isn't loaded. Otherwise this could have some side effects if you don't create the doctrine mapping

src/DependencyInjection/SonataClassificationExtension.php
```diff
@@ -107,131 +111,145 @@ public function registerDoctrineMapping(array $config)
 
         $collector = DoctrineCollector::getInstance();
 
-        $collector->addAssociation($config['class']['category'], 'mapOneToMany', [
-            'fieldName' => 'children',
-            'targetEntity' => $config['class']['category'],
-            'cascade' => [
+        $categoryOnyToManyOptions = OptionsBuilder::create()
+            ->add('fieldName', 'children')
```
core23: I wouldn't use this signature to create a mapping, because you must know the technical details.

There is an issue with a better naming approach: https://github.com/sonata-project/sonata-doctrine-extensions/issues/172

PR #514, opened 2020-03-15
Add new extension points to managers

core23:
## Subject

Added some more API methods, so you can get context related categories, collections and tags.

## Changelog

```markdown
### Added
- Added `CollectionManager::getBySlug` method
- Added `CategoryManager::getBySlug` method
- Added `CategoryManager::getByContext` method
- Added `TagManager::getBySlug` method
- Added `TagManager::getByContext` method
```

### DIFF
--- a/src/Entity/CategoryManager.php
+++ b/src/Entity/CategoryManager.php
@@ -228,6 +228,22 @@ class CategoryManager extends BaseEntityManager implements CategoryManagerInterf
         return $pager;
     }
 
+    public function getBySlug(string $slug, $context = null, ?bool $enabled = true): ?CategoryInterface
+    {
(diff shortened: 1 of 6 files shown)


greg0ire:
The added code looks good, the existing code is classes suffixed with `Manager`, that you propose to use as repositories, and this PR makes existing interfaces grow rather than adding new ones. Do you think there is a risk that even more methods are added to these classes and interfaces in the future? If yes, how do we prevent that?

core23:
> Do you think there is a risk that even more methods are added to these classes and interfaces in the future? If yes, how do we prevent that?

I don't think that our interfaces will grow any larger. In the last years there wasn't any request to add new methods.

PR #516, opened 2020-03-16
Add admin filters

core23:
## Subject

Adds reuseable filters that can respect the admins classification context.

I am targeting this branch, because this feature is BC.

## Changelog

```markdown
### Added
- Added `CategoryFilter` for admin lists
- Added `CollectionFilter` for admin lists
```

### DIFF
--- a/src/Admin/Filter/CategoryFilter.php
+++ b/src/Admin/Filter/CategoryFilter.php
@@ -1,0 +1,125 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Admin\Filter;
+
+use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
+use Sonata\ClassificationBundle\Model\CategoryInterface;
+use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
+use Sonata\DoctrineORMAdminBundle\Filter\Filter;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+
+final class CategoryFilter extends Filter
+{
+    /**
+     * @var CategoryManagerInterface
+     */
+    private $categoryManager;
+
+    public function __construct(CategoryManagerInterface $categoryManager)
+    {
+        $this->categoryManager = $categoryManager;
+    }
+
+    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data): void
+    {
+        if (null === $data || !\is_array($data) || !\array_key_exists('value', $data)) {
+            return;
+        }
+
+        if (null !== $data['value']) {
+            $queryBuilder
+                ->andWhere(sprintf('%s.%s = :category', $alias, $field))
+                ->setParameter('category', $data['value'])
+            ;
+        }
+
+        $this->active = null !== $data['value'];
+    }
+
+    public function getDefaultOptions(): array
+    {
+        return [
+            'context' => null,
+        ];
+    }
+
+    public function getFieldType(): string
+    {
+        return $this->getOption('field_type', ChoiceType::class);
+    }
+
+    public function getFieldOptions(): array
+    {
+        return $this->getOption('choices', [
+            'choices' => $this->getChoices(),
+            'choice_translation_domain' => false,
+        ]);
+    }
+
+    public function getRenderSettings(): array
+    {
+        return [DefaultType::class, [
+            'field_type' => $this->getFieldType(),
+            'field_options' => $this->getFieldOptions(),
+            'label' => $this->getLabel(),
+        ]];
+    }
+
+    protected function association(ProxyQueryInterface $queryBuilder, $data): array
+    {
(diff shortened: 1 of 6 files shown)


src/Admin/Filter/CategoryFilter.php
```diff
@@ -0,0 +1,136 @@
...
+
+    public function getFieldOptions()
+    {
+        return $this->getOption('choices', [
+            'choices' => $this->getChoices(),
+            'choice_translation_domain' => false,
+        ]);
+    }
+
+    public function getRenderSettings()
```
greg0ire: Please add return type hints, and for the other class too

src/Admin/Filter/CategoryFilter.php
```diff
@@ -0,0 +1,136 @@
...
+            $queryBuilder
+                ->andWhere(sprintf('%s.%s = :category', $alias, $field))
+                ->setParameter('category', $data['value'])
+            ;
+        }
+
+        $this->active = null !== $data['value'];
+    }
+
+    public function getDefaultOptions(): array
```
OskarStark:
```suggestion
    /**
    * @return array<string, null>
    */
    public function getDefaultOptions(): array
```

Shall we add those for further static code analysis support?

src/Admin/Filter/CollectionFilter.php
```diff
@@ -0,0 +1,124 @@
...
+
+    public function __construct(CollectionManagerInterface $collectionManager)
+    {
+        $this->collectionManager = $collectionManager;
+    }
+
+    /**
+     * @param string            $alias
+     * @param string            $field
+     * @param array|string|null $data
```
VincentLanglet, line 38: I see everywhere in these bundle checks like is_array for the $data value.

But according the interface, it's always an array
image

src/Admin/Filter/CategoryFilter.php
```diff
@@ -0,0 +1,125 @@
...
+        $this->categoryManager = $categoryManager;
+    }
+
+    public function filter(ProxyQueryInterface $queryBuilder, $alias, $field, $data): void
+    {
+        if (null === $data || !\is_array($data) || !\array_key_exists('value', $data)) {
+            return;
+        }
+
+        if ($data['value']) {
```
phansys, line 41: Could not `$data['value']` hold `null`, `0`, `false` or `''`?

tests/Admin/Filter/QueryBuilder.php
```diff
@@ -0,0 +1,158 @@
...
+    public function setParameter($key, $value, $type = null)
+    {
+        $this->parameters[$key] = $value;
+    }
+
+    public function andWhere()
+    {
+        $query = \func_get_args();
+
+        $this->query[] = (string) $query;
```
phansys, line 42: Isn't this lead to an "array to string conversion" error?

src/Admin/Filter/CategoryFilter.php
```diff
@@ -0,0 +1,125 @@
...
+        foreach ($categories as $category) {
+            $choices[sprintf('%s (%s)', $category->getName(), $category->getContext()->getId())] = $category->getId();
+
+            $this->childWalker($category, $choices);
+        }
+
+        return $choices;
+    }
+
+    private function childWalker(CategoryInterface $category, array &$choices, int $level = 2): void
```
greg0ire, line 113: naming a method as a noun is unexpected, it's not even an accessor. How about `visitChild` instead

tests/Admin/Filter/CollectionFilterTest.php
```diff
@@ -0,0 +1,47 @@
...
+class CollectionFilterTest extends TestCase
+{
+    /**
+     * @var MockObject&CollectionManagerInterface
+     */
+    private $collectionManager;
+
+    protected function setUp(): void
+    {
+        $this->collectionManager = $this->createMock(CollectionManagerInterface::class);
```
greg0ire, line 31:
```suggestion
        $this->collectionManager = $this->createStub(CollectionManagerInterface::class);
```

tests/Admin/Filter/QueryBuilder.php
```diff
@@ -0,0 +1,158 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Tests\Admin\Filter;
+
+use Doctrine\ORM\Query\Expr\Andx;
+use Doctrine\ORM\Query\Expr\Orx;
+
+final class QueryBuilder extends \Doctrine\ORM\QueryBuilder
```
greg0ire, line 19: Why not use return type hints in this class?

tests/Admin/Filter/QueryBuilder.php
core23: It's basically a copy & paste of https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/tests/Filter/QueryBuilder.php

phansys: I don't see where we are casting the result of `func_get_args()` as string at sonata-project/doctrine-orm-admin-bundle.
Just for the records, here is the example: https://3v4l.org/6UZbV.

core23:
Please review @sonata-project/contributors

src/Admin/Filter/CollectionFilter.php
```diff
@@ -0,0 +1,111 @@
...
+
+        if (null === $context) {
+            $collections = $this->collectionManager->findAll();
+        } else {
+            $collections = $this->collectionManager->getByContext($context);
+        }
+
+        $choices = [];
+
+        /** @var CollectionInterface $collection */
```
greg0ire, line 104: Why is this needed?

PR #517, opened 2020-03-18
Add manager aliases

core23:
## Subject

Add public aliases to all manager interface so that they can be used with DI.

## Changelog

```markdown
### Added
- Add public aliases to all manager interface
```

### DIFF
--- a/src/Resources/config/orm.xml
+++ b/src/Resources/config/orm.xml
@@ -24,5 +24,9 @@
             <argument>%sonata.classification.manager.context.entity%</argument>
             <argument type="service" id="doctrine"/>
         </service>
+        <service id="Sonata\ClassificationBundle\Model\CategoryManagerInterface" alias="sonata.classification.manager.category" public="true"/>
+        <service id="Sonata\ClassificationBundle\Model\TagManagerInterface" alias="sonata.classification.manager.tag" public="true"/>
+        <service id="Sonata\ClassificationBundle\Model\CollectionManagerInterface" alias="sonata.classification.manager.collection" public="true"/>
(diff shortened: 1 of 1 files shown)


VincentLanglet:
I haven't declared a service in a long time, but for what I remembered they recommend class as id.

Shouldn't we start using the new service declaration syntax `Sonata\ClassificationBundle\Model\ContextManager` instead of
`sonata.classification.manager.context` for all the Sonata services ?

core23:
> Shouldn't we start using the new service declaration syntax `Sonata\ClassificationBundle\Model\ContextManager` instead of
> `sonata.classification.manager.context` for all the Sonata services ?

We can't change this with a minor release, but we should not named ids with the next major(s).

VincentLanglet:
> We can't change this with a minor release, but we should not named ids with the next major(s).

I agree. Since the 4.x is already hard to get, should we plan this for 5.x instead ?
What's the best to plan things for all the project ? Should I create an issue in every project ?

core23:
I would not plan to far. We try to release new major for nearly two years...

VincentLanglet:
> I would not plan to far. We try to release new major for nearly two years...

I don't understand. That's why I try to not add more work in the 4.x release and was thinking that this not necessary change could wait for the next one.
And I'll hope we'll succeed in making major release at least every year after the 4.x 😄

core23:
> I don't understand. That's why I try to not add more work in the 4.x release and was thinking that this not necessary change could wait for the next one.

Let's move this discussion to slack. Hopefully I can help with the next steps on Friday when I got my day off.

PR #488, opened 2019-12-25
Remove block deprecations

core23:
## Subject

I am targeting this branch, because this is BC.

## Changelog

```markdown
### Removed
- Remove block deprecations
```

PR #522, opened 2020-03-30
Add support for BlockBundle 4.x

wbloszyk:
## Subject

@core23 
 After check #488 I think we can add support for BlockBundle 4.x. Can u check it?

I am targeting this branch, because it is PATCH.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "php": "^7.1",
         "cocur/slugify": "^2.0 || ^3.0 || ^4.0",
-        "sonata-project/admin-bundle": "^3.31",
+        "sonata-project/admin-bundle": "^3.31 || dev-master",
         "sonata-project/core-bundle": "^3.13",
         "sonata-project/datagrid-bundle": "^2.3",
         "sonata-project/doctrine-extensions": "^1.6.0",
--- a/composer.json
+++ b/composer.json
@@ -41,13 +41,13 @@
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.1 || >=3.0",
         "jms/serializer": "<0.13",
-        "sonata-project/block-bundle": "<3.14 || >=4.0",
+        "sonata-project/block-bundle": "<3.14",
         "sonata-project/media-bundle": "<3.17 || >=4.0"
     },
     "require-dev": {
         "friendsofsymfony/rest-bundle": "^2.1",
         "jms/serializer-bundle": "^1.0 || ^2.0",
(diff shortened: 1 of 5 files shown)


wbloszyk:
@core23 What do you think about it? Can we remove AbstractAdminBlockService? 
It is not used in any place. You can find only method call like here:
https://github.com/sonata-project/SonataPageBundle/blob/master/src/Admin/BlockAdmin.php#L217-L238

@sonata-project/contribiutors
If we can do it like BC, then major upgrade for other sonata-project bundles will be much simpler.

src/Block/Service/AbstractClassificationBlockService.php
```diff
@@ -56,6 +69,98 @@ public function __construct(
...
+        $this->buildEditForm($formMapper, $block);
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method.
+     *
+     * @deprecated since sonata-project/block-bundle 3.16, to be removed in version 4.0.
+     */
+    public function buildEditForm(FormMapper $form, BlockInterface $block)
+    {
```
core23:
```suggestion
    {
         $this->buildEditForm($formMapper, $block);
```

core23:
Looking at Travis, sonata-block 4 is never resolved

src/Block/Service/AbstractClassificationBlockService.php
wbloszyk: This is AbstractAdminBlockService method, we should not change then. Self reference method make infinity loop.

wbloszyk:
@core23 
I think other dependencies should be update first (like MediaBundle). What do you think about this changes? Should I work on this?

core23:
> @core23
> I think other dependencies should be update first (like MediaBundle). What do you think about this changes? Should I work on this?

As said several time on many different issues. We had some major changes on the old core bundle and the block bundle. This changes can only be resolved with new major versions on our projects. 

The main goal is still a new admin bundle major version. After that we can try to find a way for all other projects.

composer.json
```diff
@@ -23,8 +23,8 @@
     "require": {
         "php": "^7.1",
         "cocur/slugify": "^2.0 || ^3.0 || ^4.0",
-        "sonata-project/admin-bundle": "^3.31",
-        "sonata-project/core-bundle": "^3.14",
+        "sonata-project/admin-bundle": "^3.31 || dev-master",
```
core23, line 26: We should not add this to the stable branch

core23 (changes requested):

PR #531, opened 2020-05-20
Make CoreBundle optional

wbloszyk:
## Subject

I am targeting this branch, because this changes are BC.

Part of https://github.com/sonata-project/dev-kit/issues/697

## Changelog

```markdown
### Removed
- SonataCoreBundle dependencies
- Support for Symfony < 4.3
```


jordisala1991:
Why you closed this one?

wbloszyk:
Change in this bundle will be more harder. I focus on admin and block first. This will be reopen soon.

src/Controller/Api/TagController.php
```diff
@@ -238,7 +238,9 @@ protected function handleWriteTag($request, $id = null)
             'csrf_protection' => false,
         ]);
 
-        FormHelper::removeFields($request->request->all(), $form);
+        if (class_exists(FormHelper::class)) {
```
jordisala1991, line 241: I think we need to find a solution for this.. @core23 you know what is the use for this remove?

wbloszyk:
```php
    /**
     * This function remove fields available if there are not present in the $data array
     * The data array might come from $request->request->all().
     *
     * This can be usefull if you don't want to send all fields will building an api. As missing
     * fields will be threated like null values.
     */
    public static function removeFields(array $data, Form $form)
    {
        $diff = array_diff(array_keys($form->all()), array_keys($data));

        foreach ($diff as $key) {
            $form->remove($key);
        }

        foreach ($data as $name => $value) {
            if (!\is_array($value)) {
                continue;
            }

            self::removeFields($value, $form[$name]);
        }
    }
```

wbloszyk: Should I copy it to this controllers. WDYT?

jordisala1991: How many bundles use those methods?

wbloszyk: Only this.

jordisala1991: What about create the class (final) with only that method?

jordisala1991:
prefer-lowest build is failing, maybe you can try upgrading a bit those dependencies trying to find which one makes everything slow

jordisala1991:
Maybe remove sf 3.4 and leave only 4.3 could help

jordisala1991:
Thank you @wbloszyk

PR #540, opened 2020-06-21
3.12.1

greg0ire:


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.12.1](https://github.com/sonata-project/SonataClassificationBundle/compare/3.12.0...3.12.1) - 2020-06-21
+### Fixed
+- [[#539](https://github.com/sonata-project/SonataClassificationBundle/pull/539)]
+  Fix mysql database schema ([@wbloszyk](https://github.com/wbloszyk))
```
greg0ire, line 8: @jordisala1991 maybe we should remove the links? Github does them automatically…

jordisala1991: On the releases page it shows automatically links to users and issues/pr. But on the changelog file that does not happen automatically. Should we remove them anyway?

greg0ire: Maybe we could output a changelog with links to stdout, and one without to stderr? That way we copy-paste from the terminal when creating the release. I'm not sure people get notified when used in a link like that…

greg0ire: Or maybe, no links to stderr, no handles or PR numbers either, because the CHANGELOG should not be tied to Github.
And for stdout, handles and PR numbers but as links?

jordisala1991: I think we should show on both sides the PR link and the author link, but on the release we can remove the links because github can generate them automatically.

On the other side, I think we should automate that thing, and the way we output to stdout or stderr is only a workaround because we are not there yet. It is a first step.

VincentLanglet: I think it's nice to add the author and the PR link. It helps to find the related PR if I think it introduce a bug or if I want to know more about the changes.

greg0ire: Ok I will leave the link markup in the changelog, and remove it from the release (while still leaving the text for the link, which should be resolved by GH)

PR #580, opened 2020-09-06
Manual merge

franmomu:


franmomu:
There is no need of manual merge.

PR #583, opened 2020-09-15
Fix upgrade note

wbloszyk:


phansys (approved):
IMO, this note is correct as is.
This change is only required to keep the message identical to the upgrade notes in other packages.

jordisala1991 (approved):

jordisala1991:
Thank you @wbloszyk

PR #678, opened 2021-06-15
Add compatibility with Sonata4 and Symfony5

VincentLanglet:
## Changelog

```markdown
### Added
- Added compatibility with Symfony 5
- Added compatibility with Twig 3
```

PR #680, opened 2021-06-18
Inject PsrContainer instead of full container

VincentLanglet:
Ref: https://github.com/sonata-project/SonataClassificationBundle/pull/678#discussion_r654155892

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,8 @@
         "doctrine/mongodb-odm": "^2.0",
         "friendsofsymfony/rest-bundle": "^2.3 || ^3.0",
         "jms/serializer-bundle": "^2.0 || ^3.0",
+        "knplabs/knp-menu": "^3.0",
+        "knplabs/knp-menu-bundle": "^3.0",
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
         "nelmio/api-doc-bundle": "^2.13.4 || ^3.6",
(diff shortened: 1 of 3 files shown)


franmomu:
based on the error, I guess `SonataAdminBundle` should be registered in https://github.com/sonata-project/SonataClassificationBundle/blob/fb9c9b1c33a1c784a7e22be51c3d5256fe53a581/tests/App/AppKernel.php#L41-L54

VincentLanglet:
Thanks @franmomu, build is fixed now

composer.json
```diff
@@ -24,6 +24,8 @@
         "php": "^7.3 || ^8.0",
         "cocur/slugify": "^3.0 || ^4.0",
         "doctrine/persistence": "^1.3.6 || ^2.0",
+        "knplabs/knp-menu": "^3.0",
```
jordisala1991, line 27: Do you need this one?

composer.json
```diff
@@ -24,6 +24,8 @@
...
+        "knplabs/knp-menu-bundle": "^3.0",
```
jordisala1991, line 28: if the admin bundle is on require dev, this should be too on dev right?

composer.json
VincentLanglet: https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Resources/views/CategoryAdmin/list_tab_menu.html.twig#L1

We're using  knp_menu_render which is define in knp-menu, so it seems better to add it to the require.

composer.json
VincentLanglet: Thought it should be a dependency because it's used in the src https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Resources/views/CategoryAdmin/list_tab_menu.html.twig#L1

But I moved it to require-dev if preferred

jordisala1991 (approved):

PR #681, opened 2021-06-20
Fix some phpstan issues, add generics to repositories

jordisala1991:
## Subject

I am targeting this branch, because this breaks BC

## Changelog

```markdown
### Removed
- Compatibility with `sonata-project/datagrid-bundle` < 3.0
- Compatibility with `sonata-project/form-extensions` < 1.4
```

### DIFF
--- a/src/Controller/Api/CollectionController.php
+++ b/src/Controller/Api/CollectionController.php
@@ -223,7 +223,7 @@ class CollectionController
      *
      * @throws NotFoundHttpException
      *
-     * @return Rest\View
+     * @return View
      */
     public function deleteCollectionAction($id)
     {
@@ -231,7 +231,7 @@ class CollectionController
 
         $this->collectionManager->delete($collection);
 
-        return ['deleted' => true];
+        return View::create(['deleted' => true]);
     }
 
     /**
@@ -280,7 +280,7 @@ class CollectionController
      * @param Request     $request Symfony request
      * @param string|null $id      Collection identifier
      *
-     * @return Rest\View|FormInterface
+     * @return View|FormInterface
      */
     protected function handleWriteCollection($request, $id = null)
     {
--- a/src/Model/CategoryManagerInterface.php
+++ b/src/Model/CategoryManagerInterface.php
@@ -27,6 +27,7 @@ use Sonata\Doctrine\Model\ManagerInterface;
  * @method CategoryInterface|null getBySlug(string $slug, ContextInterface|string|null $context, bool $enabled = true)
  *
  * @phpstan-extends ManagerInterface<CategoryInterface>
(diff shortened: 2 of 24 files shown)


VincentLanglet (commented):
make psalm and make phpstan should be added to the Makefile.

Also, should we modify dev-kit in order to add phpstan/psalm on a branch instead of a project ?

jordisala1991:
> make psalm and make phpstan should be added to the Makefile.
> 
> Also, should we modify dev-kit in order to add phpstan/psalm on a branch instead of a project ?

I will se if it can be added on 3.x first

src/Controller/Api/CollectionController.php
```diff
@@ -279,13 +280,13 @@ protected function getCollection($id)
...
      * @param string|null $id      Collection identifier
      *
-     * @return Rest\View|FormInterface
+     * @return View|FormInterface
      */
     protected function handleWriteCollection($request, $id = null)
     {
         $collection = $id ? $this->getCollection($id) : null;
 
-        $form = $this->formFactory->createNamed('', 'sonata_classification_api_form_collection', $collection, [
```
VincentLanglet, line 288: If this was working, should we provide a PR to psalm in order to not restrict the param to class-string ?

jordisala1991: Symfony changed the form types since 3.4 I think. They have to be class-string. I don't think this code was working/tested enough

jordisala1991: Also, tbh, I don't think anyone is using this api, they was so much wrong with it (previously we had null on the first parameter, another thing that it does not work)

src/Model/CategoryManagerInterface.php
```diff
@@ -25,8 +25,10 @@
  * @method CategoryInterface[]    getRootCategoriesSplitByContexts(bool $loadChildren = true)
  * @method CategoryInterface[]    getCategories(ContextInterface|string|null $context)
  * @method CategoryInterface|null getBySlug(string $slug, ContextInterface|string|null $context, bool $enabled = true)
+ *
+ * @phpstan-extends ManagerInterface<CategoryInterface>
  */
-interface CategoryManagerInterface extends ManagerInterface, PageableManagerInterface
+interface CategoryManagerInterface extends ManagerInterface, PageableInterface
```
VincentLanglet, line 31: The PageableInterface has 
```
@template-covariant T of object
```
so a `phpstan-extends PageableInterface< CategoryInterface >`
could also be added (same for others interface).

jordisala1991: Done

VincentLanglet (approved):
Changelog is needed for the version dropped

jordisala1991:
Added the changelog

PR #684, opened 2021-06-20
Resolve some deprecation

VincentLanglet:
## Subject

I am targeting this branch, because BC-break.

## Changelog

```markdown
### Removed
- `CategoryManager:: getRootCategory()` use `getRootCategoriesForContext()` instead
- `CategoryManager:: getRootCategories()` use `getAllRootCategories()` instead
- `CategoryManager:: getCategories()` use `getRootCategoriesForContext()` instead
- `CategoryManagerInterface:: getRootCategory()` use `getRootCategoriesForContext()` instead
- `CategoryManagerInterface:: getRootCategories()` use `getAllRootCategories()` instead
- `CategoryManagerInterface:: getCategories()` use `getRootCategoriesForContext()` instead
```

### DIFF
--- a/src/Document/CategoryManager.php
+++ b/src/Document/CategoryManager.php
@@ -46,9 +46,6 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
         $this->categories = [];
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getRootCategoriesPager($page = 1, $limit = 25, $criteria = [])
     {
         $queryBuilder = $this->getObjectManager()->createQueryBuilder()
@@ -64,9 +61,6 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
         return $pager;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getSubCategoriesPager($categoryId, $page = 1, $limit = 25, $criteria = [])
     {
         $queryBuilder = $this->getObjectManager()->createQueryBuilder()
@@ -81,51 +75,6 @@ class CategoryManager extends BaseDocumentManager implements CategoryManagerInte
         $pager->init();
 
         return $pager;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getRootCategory($context = null)
-    {
-        $context = $this->getContext($context);
-
-        $this->loadCategories($context);
-
-        return current($this->categories[$context->getId()]);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getRootCategories($loadChildren = true)
-    {
-        $rootCategories = $this->getObjectManager()->createQuery(sprintf('SELECT c FROM %s c WHERE c.parent IS NULL', $this->getClass()))
-            ->execute();
-
-        $categories = [];
-
-        foreach ($rootCategories as $category) {
-            if (null === $category->getContext()) {
-                throw new \RuntimeException('Context cannot be null');
-            }
-
-            $categories[$category->getContext()->getId()] = $loadChildren ? $this->getRootCategory($category->getContext()) : $category;
-        }
-
-        return $categories;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getCategories($context = null)
-    {
-        $context = $this->getContext($context);
-
-        $this->loadCategories($context);
-
-        return $this->categories[$context->getId()];
     }
 
     public function getPager(array $criteria, int $page, int $limit = 10, array $sort = []): PagerInterface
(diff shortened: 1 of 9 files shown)


jordisala1991 (dismissed):

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getRootCategoryWithChildren(CategoryInterface $category): void
+    public function getRootCategoryWithChildren(CategoryInterface $category)
     {
-        throw new \RuntimeException('Not Implemented yet');
+        if (null === $category->getContext()) {
+            throw new \RuntimeException('Context cannot be null');
```
phansys, line 112:
```suggestion
            throw new \InvalidArgumentException('Context cannot be null.');
```

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+        }
+        if (null !== $category->getParent()) {
+            throw new \RuntimeException('Method can be called only for root categories');
```
phansys, line 115:
```suggestion
            throw new \InvalidArgumentException('Method can be called only for root categories.');
```

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+
+        $this->loadCategories($context);
+
+        foreach ($this->categories[$context->getId()] as $contextRootCategory) {
+            if ($category->getId() === $contextRootCategory->getId()) {
+                return $contextRootCategory;
+            }
+        }
+
+        throw new \RuntimeException('Category does not exist');
```
phansys, line 128:
```suggestion
        throw new \InvalidArgumentException('Category does not exist.');
```
BTW, is there a identifier we could use for the referenced category in these exception messages? Something like this:
>Category "ref" does not exist.

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+        $class = $this->getClass();
+
+        $rootCategories = $this->getObjectManager()->createQuery(sprintf('SELECT c FROM %s c WHERE c.parent IS NULL', $class))
+            ->execute();
+
+        $categories = [];
+
+        foreach ($rootCategories as $category) {
+            if (null === $category->getContext()) {
+                throw new \RuntimeException('Context cannot be null');
```
phansys, line 151:
```suggestion
                throw new \LogicException('Context cannot be null.');
```

phansys (changes requested):

VincentLanglet:
Done @phansys

jordisala1991 (approved):

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
-    /**
-     * {@inheritdoc}
-     */
-    public function getAllRootCategories($loadChildren = true): void
+    public function getAllRootCategories($loadChildren = true)
     {
-        throw new \RuntimeException('Not Implemented yet');
+        $class = $this->getClass();
+
+        $rootCategories = $this->getObjectManager()->createQuery(sprintf('SELECT c FROM %s c WHERE c.parent IS NULL', $class))
```
franmomu, line 144: I guess this has never been used, maybe it's better to leave it with the exception?

It is supposed to be a `DocumentManager` and there is no `DocumentManager::createQuery` method.

VincentLanglet: Didn't know, since there was already these call to `createQuery`.
https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Document/CategoryManager.php#L103
https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Document/CategoryManager.php#L201

What would be the right way to implement this ?

VincentLanglet:
Can you take a new look @phansys ?

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+            ->execute();
+
+        $categories = [];
+
+        foreach ($rootCategories as $category) {
+            if (null === $category->getContext()) {
+                throw new \LogicException('Context cannot be null');
```
phansys, line 151:
```suggestion
                throw new \LogicException('Context cannot be null.');
```

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
-    /**
-     * {@inheritdoc}
-     */
-    public function getRootCategoriesSplitByContexts($loadChildren = true): void
+    public function getRootCategoriesSplitByContexts($loadChildren = true)
     {
-        throw new \RuntimeException('Not Implemented yet');
+        $rootCategories = $this->getAllRootCategories($loadChildren);
+
+        $splitCategories = [];
```
phansys, line 164:
```suggestion
        $splittedCategories = [];
```

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+
+        foreach ($rootCategories as $category) {
+            $splitCategories[$category->getContext()->getId()][] = $category;
```
phansys, line 167:
```suggestion
            $splittedCategories[$category->getContext()->getId()][] = $category;
```

src/Document/CategoryManager.php
```diff
@@ -157,36 +106,85 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+        }
+
+        return $splitCategories;
```
phansys, line 170:
```suggestion
        return $splittedCategories;
```

src/Entity/CategoryManager.php
```diff
@@ -169,7 +127,7 @@ public function getAllRootCategories($loadChildren = true)
 
         foreach ($rootCategories as $category) {
             if (null === $category->getContext()) {
-                throw new \RuntimeException('Context cannot be null');
+                throw new \LogicException('Context cannot be null');
```
phansys, line 130:
```suggestion
                throw new \LogicException('Context cannot be null.');
```

phansys (changes requested):

src/Document/CategoryManager.php
VincentLanglet: Done

src/Document/CategoryManager.php
VincentLanglet: Done

src/Document/CategoryManager.php
VincentLanglet: Done

src/Document/CategoryManager.php
VincentLanglet: Done

src/Entity/CategoryManager.php
VincentLanglet: Done

phansys (approved):

VincentLanglet:
@phansys seems like split was correct and not splitted: https://www.englishforums.com/English/SplittedIsItCorrect/bbxlql/post.htm

phansys:
>@phansys seems like split was correct and not splitted: https://www.englishforums.com/English/SplittedIsItCorrect/bbxlql/post.htm

Oh, that seems correct. I didn't notice it can be used as adjective too. I understood it was used as an imperative verb in this case.
Thank you for pointing.

phansys (dismissed):

src/Document/CategoryManager.php
franmomu: Yeah, I meant that has never used because it is the same code it was before. 

I've never used this bundle, so I have no idea of its models and associations, so I'd rather throw an exception for now instead of adding some new untested code (that could be added in `3.x`).

VincentLanglet:
Should be ok now @franmomu

src/Document/CategoryManager.php
```diff
@@ -157,36 +108,87 @@ public function getPager(array $criteria, int $page, int $limit = 10, array $sor
...
+            ->createQueryBuilder('c')
+            ->where('c.parent IS NULL')
+            ->getQuery()
+            ->getResult();
+
+        $categories = [];
+
+        foreach ($rootCategories as $category) {
+            if (null === $category->getContext()) {
+                throw new \LogicException('Context cannot be null.');
```
phansys, line 155: IMO, this message is misleading since there is no way to know which category is missing the context.

VincentLanglet: Done

phansys (approved):

jordisala1991 (approved):

PR #721, opened 2021-09-07
Bump PHPStan level to 2

phansys:
## Subject
Bump PHPStan level to 2.

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Added
- Missing methods in interfaces (through the `@method` annotation).

### Fixed
- Calls to several undefined methods.
```

## To do

- [x] Validate the changes related to the ODM manager (/cc @franmomu);

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,24 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### `Sonata\ClassificationBundle\Model\CategoryInterface`
+
+The classes implementing the interface `CategoryInterface` SHOULD declare the methods
+`__toString()` and `getId()`, as they will be required in version 4.x.
+
+### `Sonata\ClassificationBundle\Model\CollectionInterface`
+
+The classes implementing the interface `CollectionInterface` SHOULD declare the method
+`getId()`, as it will be required in version 4.x.
+
+### `Sonata\ClassificationBundle\Model\TagInterface`
+
+The classes implementing the interface `TagInterface` SHOULD declare the method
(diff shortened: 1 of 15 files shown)


src/Document/CategoryManager.php
```diff
@@ -24,7 +24,7 @@ public function getPager(array $criteria, $page, $limit = 10, array $sort = [])
     {
         $parameters = [];
 
-        $query = $this->getRepository()
+        $query = $this->getDocumentManager()
```
franmomu, line 27: I think this would be something like:

```php
$query = $this->getDocumentManager()->createQueryBuilder();
$query->field('enabled')->equals($criteria['enabled'] ?? true);

$pager = new Pager();
$pager->setMaxPerPage($limit);
$pager->setQuery(new ProxyQuery($query));
// ...
```

phansys: I added a new commit @franmomu: 9b188f5415668d14ad890ae4edc34dd3773f3eee.
Could you please check and confirm if this seems fine to you?
Is there any way this ever worked the way it was written before?
If no, I think we must provide a separate PR for these changes.
Thank you.

franmomu: Yeah, this looks fine and I would say it has never worked before, those methods did not exist in the odm 😕

VincentLanglet (dismissed):

VincentLanglet (approved):

src/Document/CategoryManager.php
phansys: Fixed in #724.
Thank you so much.

franmomu:
Thanks @phansys !

PR #745, opened 2021-10-13
Add type hints to methods and properties

core23:
## Subject

I am targeting this branch, because this is a BC break.

## Changelog

```markdown
### Changed
- Added type hints to methods and properties
- Reduce visibility of config methods

### Removed
- Removed `Sonata\ClassificationBundle\Form\FormHelper`
```


VincentLanglet (dismissed):

src/Entity/CategoryManager.php
```diff
@@ -66,7 +60,7 @@ public function getRootCategoriesPager($page = 1, $limit = 25, $criteria = [])
         return $pager;
     }
 
-    public function getSubCategoriesPager($categoryId, $page = 1, $limit = 25, $criteria = [])
+    public function getSubCategoriesPager(int $categoryId, int $page = 1, int $limit = 25, array $criteria = []): Pager
```
wbloszyk: IMO we should not define `catrgoryId` type and aby other id field.

wbloszyk (changes requested):

src/Entity/CategoryManager.php
ChristianGripp: Fixed

wbloszyk (dismissed):

Makefile
```diff
@@ -45,6 +45,7 @@ lint-xliff:
 
 lint-php:
 	php-cs-fixer fix --ansi --verbose --diff --dry-run
+
```
VincentLanglet: weird

psalm-baseline.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
+    <file src="src/Block/Service/AbstractClassificationBlockService.php">
```
VincentLanglet: What is the issue with all of this ? Can't we fix those ?

Makefile
ChristianGripp: Indeed. The last change on the 4.x build successfully

VincentLanglet (commented):
Can't we fix the baseline instead of generating it ?

psalm-baseline.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
+    <file src="src/Entity/CollectionManager.php">
+        <QueryBuilderSetParameter occurrences="2">
```
ChristianGripp: Not sure how to fix this. The context can be a string type or a context instance.

VincentLanglet: Do you know where/how are used the getBySlug and getByContext method ?

First, It could be great to have a method only accepting string or Context instance but not both. But that shouldn't be a general blocker but it might have sens here.

If I remember correctly calling `setParameter` on an object is kinda the same than calling setParameter and passing the id of this object. Which means that the string that can be passed to this method must be the id of the context, not the name of the method, this can be confusing.

Anyway. Passing an object to `setParameter` without specifying a type is not recommended for perf issues. https://twitter.com/Ocramius/status/1251478780381716480

Best would be to call getId on it if it's an object.

ChristianGripp: > First, It could be great to have a method only accepting string or Context instance but not both. But that shouldn't be a general blocker but it might have sens here.

Did not find any usage for a context object parameter, so I changed this to string only.

VincentLanglet: Then we can rename it contextId

ChristianGripp: > Then we can rename it contextId

We should keep the name as it is. We  use `context` across all bundles when talking about the string representation.

VincentLanglet: Do you have some example ?

Because when I look at the code in the ClassificationBundle I see code like
```
/* @var ContextInterface $context */
foreach ($this->contextManager->findAll() as $context) {
    $contextChoices[(string) $context->getId()] = (string) $context->getName();
}
```

So it doesn't seem a bad thing to use `contextId` for the string representation. And it's not because the mistake was made somewhere else, that we should keep it forever. The 4.0 release might be the right time to fix it everywhere.

Fact is I don't use other bundle, and I don't know what is this context. I looked at the MediaBundle to ask the opinion of @jordisala1991 but seems like all the `string $context` usage are related to a real string value. A media has a context property, not a ContextInterface one.

ChristianGripp: > Do you have some example ?

When searching for `context` inside the media bundle, there are several findings:
https://github.com/sonata-project/SonataMediaBundle/blob/26991351cac2b1399890685b43c4d20efc69da11/src/Block/GalleryListBlockService.php#L79
https://github.com/sonata-project/SonataMediaBundle/blob/cafbe30674803dd5bfebd6f6e7cfbbf5cafb5e6d/src/Admin/GalleryAdmin.php#L69
https://github.com/sonata-project/SonataNewsBundle/blob/75b0c68e9bb77d75f0605520eb1af12c63071b36/src/Admin/PostAdmin.php#L122

There are some more, but I'm lazy to link every code fragment ;)

> A media has a context property, not a ContextInterface one.

That's right, but the context property of a media instance is linked to the context id of the `ContextInterface`:

https://github.com/sonata-project/SonataMediaBundle/blob/961e51a0da4b0be82ca0ada2644422f918c4e6f2/src/Command/FixMediaContextCommand.php#L70-L72

core23:
Please review @sonata-project/contributors 

I'd like to release a first alpha version soon

wbloszyk (dismissed):

src/Block/Service/AbstractClassificationBlockService.php
```diff
@@ -25,13 +25,14 @@
 
 /**
  * @author Christian Gripp <mail@core23.de>
+ *
+ * @phpstan-import-type FieldDescriptionOptions from \Sonata\AdminBundle\FieldDescription\FieldDescriptionInterface
+ *
+ * @phpstan-template T of object of ContextAwareInterface
```
ChristianGripp: Looks like psalm has some problem with this :/

Can someone help or should we ignore this error?

VincentLanglet: Didnt take a big look, but I would say you want to do
```
@phpstan-template T of ContextAwareInterface
```

src/Block/Service/AbstractClassificationBlockService.php
```diff
@@ -41,15 +42,15 @@ public function __construct(Environment $twig, ContextManagerInterface $contextM
...
-     * @param string $formField
-     * @param string $field
-     * @param array  $fieldOptions
-     * @param array  $adminOptions
+     * @param array<string, mixed> $fieldOptions
+     * @param array<string, mixed> $adminOptions
      *
-     * @phpstan-param AdminInterface<ContextAwareInterface> $admin
+     * @phpstan-param FormMapper<T> $formMapper
+     * @phpstan-param AdminInterface<T> $admin
```
ChristianGripp, line 49: The admin generic requires and object and no interface:
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Admin/AbstractAdmin.php#L55

src/Block/Service/AbstractClassificationBlockService.php
ChristianGripp: This does not work (see: https://github.com/sonata-project/SonataClassificationBundle/pull/745/files/0bd9a6a6e7603b70db4608365f00c3c6e313fc54#r736730231)

VincentLanglet: The class considered is `objectofContextAwareInterface`.

It should work with `phpstan-template T of ContextAwareInterface` https://psalm.dev/r/97f4adf72f

src/Block/Service/AbstractClassificationBlockService.php
VincentLanglet: This shouldnt be an issue https://psalm.dev/r/97f4adf72f

ChristianGripp: Now we got a different error, because an interface generic is not an object generic :/

VincentLanglet: This is a "normal" error, because phpstan considered that the Admin must support any object, because we used `AdminInterface<object>`.

We solve this by making the whole class generic, or sometimes juste making the method kinda-generic https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Admin/AdminHelper.php#L87-L91

I think that chanigng setAdmin and setAssociationAdmin the same way could solve those errors.

ChristianGripp: This should fix it: https://github.com/sonata-project/SonataAdminBundle/pull/7558

VincentLanglet (dismissed):

psalm.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="8" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd">
+<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="6" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" errorBaseline="psalm-baseline.xml">
```
phansys: Is it fine to reduce the introspection level?

tests/Entity/TagManagerTest.php
```diff
@@ -60,7 +60,7 @@ public function testGetByContext(): void
             ->getByContext('contextA', false);
     }
 
-    private function getTagManager($qbCallback): TagManager
+    private function getTagManager(\Closure $qbCallback): TagManager
```
phansys, line 63: Do we need to force the passed callable to be anonymous? Otherwise, I think we should use `callable` instead.

psalm.xml
VincentLanglet: In psalm 6 is stricter than 8

phansys: :baby_chick:

tests/Entity/TagManagerTest.php
ChristianGripp: The `createEntityManagerMock` method requires a closure parameter

psalm.xml
phansys: Thank you for the clarification.

phansys (dismissed):

core23:
Bumped phpstan to from level 4 to 6 and psalm down from 8 to 4.

IMHO the rest is nice to have and should not block a new release.

VincentLanglet:
> Bumped phpstan to from level 4 to 6 and psalm down from 8 to 4.
> 
> IMHO the rest is nice to have and should not block a new release.

Let's merge this and I'll take quickly a look at what's missing to bump the static analysis.

VincentLanglet (commented):
Why do we provide `getRootCategoriesPager` and `getSubCategoriesPager` in the interface if we don't use them ?

I think those method should be removed.

VincentLanglet:
Also, all the public/protected method from abstract classes can be made final (but can be made in another PR)

VincentLanglet:
The error
```
Parameter #1 $formMapper of method Sonata\ClassificationBundle\Block\Service\AbstractClassificationBlockService<Sonata\ClassificationBundle\Model\TagInterface>::getFormAdminType() expects  
         Sonata\AdminBundle\Form\FormMapper<Sonata\ClassificationBundle\Model\TagInterface>, Sonata\BlockBundle\Form\Mapper\FormMapper given.
```
is interesting. The typehint accepts `Sonata\BlockBundle\Form\Mapper\FormMapper` but we're using it like it's the AdminBundle interface.

Making EditableBlockService a generic
```
/**
 * @author Christian Gripp <mail@core23.de>
 *         
 * @phpstan-template T of FormMapper
 */
interface EditableBlockService
{
    /** @phpstan-param T $form */
    public function configureEditForm(FormMapper $form, BlockInterface $block): void;

    /** @phpstan-param T $form */
    public function configureCreateForm(FormMapper $form, BlockInterface $block): void;

    public function validate(ErrorElement $errorElement, BlockInterface $block): void;

    public function getMetadata(): MetadataInterface;
}
```

```
@phpstan-implements EditableBlockService<\Sonata\AdminBundle\Form\FormMapper<TagInterface>>
```
on AbstractTagsBlockService solve the issue.

VincentLanglet:
There is a call `$category->disableChildrenLazyLoading();` which doesn't exist on the interface ; and we need to solve this.

I made a commit bumping phpstan to 8 and psalm to 3. There is still some errors to fix but lot of them are done.
https://github.com/VincentLanglet/SonataClassificationBundle/commit/322a17efa0e4c8032520265070b9828ee418f3a8

core23:
> Why do we provide `getRootCategoriesPager` and `getSubCategoriesPager` in the interface if we don't use them ?


> Also, all the public/protected method from abstract classes can be made final (but can be made in another PR)


Not related to this PR

> The typehint accepts `Sonata\BlockBundle\Form\Mapper\FormMapper` but we're using it like it's the AdminBundle interface.

This is indeed a problem. I'll try to fix this with the next PR.

VincentLanglet:
> > Why do we provide `getRootCategoriesPager` and `getSubCategoriesPager` in the interface if we don't use them ?
> 
> > Also, all the public/protected method from abstract classes can be made final (but can be made in another PR)
> 
> Not related to this PR
> 
> > The typehint accepts `Sonata\BlockBundle\Form\Mapper\FormMapper` but we're using it like it's the AdminBundle interface.
> 
> This is indeed a problem. I'll try to fix this with the next PR.

Sure I was just using the PR to list the to do for the final 4.0 release.

VincentLanglet (approved):

PR #746, opened 2021-10-16
Extract build tools using composer-bin plugin

core23:
## Subject

Try to fix the build.

This will increase the build time as two `composer.json` projects will be installed with every build job.

Refs: https://github.com/sonata-project/dev-kit/issues/372

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ phpunit.xml
 phpstan.neon
 /.phpunit.result.cache
 /docs/_build/
+!/vendor-bin/tools/composer.lock
+/vendor-bin/tools/vendor/
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
         "sonata-project/doctrine-orm-admin-bundle": "<=4.0.0-rc.2"
     },
     "require-dev": {
+        "bamarni/composer-bin-plugin": "^1.3",
         "doctrine/annotations": "^1.7",
         "doctrine/mongodb-odm": "^2.2",
         "doctrine/orm": "^2.7.5",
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -11,3 +11,5 @@ phpunit.xml
 phpstan.neon
 /.phpunit.result.cache
 /docs/_build/
+!/vendor-bin/tools/composer.lock
```
ChristianGripp, line 14: When using composer-bin, we should manage all `vendor-bin` files (including the LOCK file) via the dev-kit.

vendor-bin/tools/composer.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "sonata-project/dev-tools",
+    "type": "project",
+    "description": "Development tools that do not conflict the project dependencies",
+    "require-dev": {
+        "friendsofphp/php-cs-fixer": "^3.0",
+        "phpspec/prophecy-phpunit": "^2.0",
```
VincentLanglet, line 7: We got rid of prophecy and prefer to use mock. We should not add it again IMHO.

vendor-bin/tools/composer.json
```diff
@@ -0,0 +1,23 @@
...
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^0.12.84",
+        "phpstan/phpstan-phpunit": "^0.12.16",
+        "phpstan/phpstan-symfony": "^0.12.21",
+        "phpunit/phpunit": "^8.5 || ^9.0",
```
VincentLanglet, line 12: Do we need to keep ^8 support ?

vendor-bin/tools/composer.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "sonata-project/dev-tools",
+    "type": "project",
+    "description": "Development tools that do not conflict the project dependencies",
+    "require-dev": {
+        "friendsofphp/php-cs-fixer": "^3.0",
```
VincentLanglet, line 6: This is a good things since it will allow to use this in the github actions

composer.json
```diff
@@ -91,10 +84,10 @@
     },
     "scripts": {
         "post-install-cmd": [
-            "[ $COMPOSER_DEV_MODE -eq 0 ] || vendor/bin/simple-phpunit install"
+            "@composer bin all install --ansi"
```
VincentLanglet, line 87: `[ $COMPOSER_DEV_MODE -eq 0 ] ||`

Was allowing to only run this in dev mode ; does bin composer do the same ?

.gitignore
VincentLanglet: You mean dispatching `vendor-bin/tools/composer.json` to all the others projects ?
Might be interesting, but then we'll install some useless phpstan extension like `phpstan/phpstan-doctrine` on non doctrine related projects.

VincentLanglet (commented):
This is an interesting solution

I don't think the lock file should be committed. It would require us to bump regularly the tool version (especially phpstan/psalm) and fix new errors for every projects. This would require a lot of extra work.
Without commiting the lock files, it will be bumped at every test run and we just have to fix when a new error happen. It happens some times, never blocked us and we avoided to do something every other times the dependency got a new version.

.gitignore
jordisala1991: If we do this direction (which gives me mixed feelings) this file should be dispatched, otherwise we need to maintain 2 composer json on each project which is a not good idea imho.

What I like about it is that we can handle all the tooling from dev kit and does not depend on each project having this correctly configured.

What I dont like is that we add more dependencies on this composer bin tool, when this can also be fixed by just installing phpunit on the project composer.json directly (in fact phpunit is a direct dev dependency used a lot in tests folder, which is not the same case for phpstan for example)

Not sure about the composer.lock either, It could be a good idea if we have some dependabot upgrading this file on dev-kit or a bad idea if we have to manually do it.

VincentLanglet: > in fact phpunit is a direct dev dependency used a lot in tests folder

And symfony decided to promote requiring both phpunit and phpunit bridge for now.

OskarStark: We can "build" the composer.json file in dev kit per project with the needed extensions/libs etc. and dispatch it

jordisala1991:
This solution is not needed anymore (for now).

PR #754, opened 2021-10-25
Next major preparations

core23:
I am targeting this branch, because this is not BC.

## Changelog

```markdown
### Added
- Add type hints to manager methods
```

### DIFF
--- a/src/Document/CategoryManager.php
+++ b/src/Document/CategoryManager.php
@@ -147,7 +147,7 @@ final class CategoryManager extends BaseDocumentManager implements CategoryManag
         return $splitCategories;
     }
 
-    public function getBySlug(string $slug, $context = null, ?bool $enabled = true): ?CategoryInterface
+    public function getBySlug(string $slug, ?string $context = null, ?bool $enabled = true): ?CategoryInterface
     {
         $queryBuilder = $this->getDocumentManager()
             ->createQueryBuilder($this->getClass())
--- a/src/Document/CategoryManager.php
+++ b/src/Document/CategoryManager.php
@@ -147,17 +147,17 @@ final class CategoryManager extends BaseDocumentManager implements CategoryManag
         return $splitCategories;
     }
 
-    public function getBySlug(string $slug, ?string $context = null, ?bool $enabled = true): ?CategoryInterface
+    public function getBySlug(string $slug, ?string $contextId = null, ?bool $enabled = true): ?CategoryInterface
     {
         $queryBuilder = $this->getDocumentManager()
             ->createQueryBuilder($this->getClass())
             ->field('slug')
             ->equals($slug);
 
-        if (null !== $context) {
+        if (null !== $contextId) {
             $queryBuilder
                 ->field('context')
-                ->equals($context);
+                ->equals($contextId);
         }
         if (null !== $enabled) {
             $queryBuilder
--- a/src/Entity/TagManager.php
+++ b/src/Entity/TagManager.php
@@ -19,15 +19,15 @@ use Sonata\Doctrine\Entity\BaseEntityManager;
 
 final class TagManager extends BaseEntityManager implements TagManagerInterface
 {
-    public function getBySlug(string $slug, ?string $context = null, ?bool $enabled = true): ?TagInterface
+    public function getBySlug(string $slug, ?string $contextId = null, ?bool $enabled = true): ?TagInterface
     {
         $queryBuilder = $this->getRepository()
             ->createQueryBuilder('t')
             ->select('t')
             ->andWhere('t.slug = :slug')->setParameter('slug', $slug);
 
-        if (null !== $context) {
-            $queryBuilder->andWhere('t.context = :context')->setParameter('context', $context);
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -45,6 +45,7 @@ lint-xliff:
 
 lint-php:
 	php-cs-fixer fix --ansi --verbose --diff --dry-run
+
```
VincentLanglet: Why this line is added ?

ChristianGripp: IMHO this was a merge conflict

src/Document/CategoryManager.php
```diff
@@ -147,7 +147,7 @@ public function getRootCategoriesSplitByContexts($loadChildren = true)
         return $splitCategories;
     }
 
-    public function getBySlug(string $slug, $context = null, ?bool $enabled = true): ?CategoryInterface
+    public function getBySlug(string $slug, ?string $context = null, ?bool $enabled = true): ?CategoryInterface
```
wbloszyk: `ContextInterface|string|null`

IMO reduce this to `?string` is good idea. People can use `$context->getId()` from `ContextInterface`. I have only worries about understanding. Context can look like:
```
'id' => 'some.context`,
'name' => 'Some context',
``` 
Can you change `$context` to `$contextId`? It will be more obvius.

wbloszyk: Second worries is what happend when `Context` is not exists?
`?string` is more flexible but `?ContextInterface` is more predictible to maintanier.

@sonata-project/contributors WDYT?

VincentLanglet: We already had the discussion somewhere else, string was preferred to ContextInterface.

I was also in favor of contextId, but @core23 was against.

ChristianGripp: > When searching for `context` inside the media bundle, there are several findings: https://github.com/sonata-project/SonataMediaBundle/blob/26991351cac2b1399890685b43c4d20efc69da11/src/Block/GalleryListBlockService.php#L79 
https://github.com/sonata-project/SonataMediaBundle/blob/cafbe30674803dd5bfebd6f6e7cfbbf5cafb5e6d/src/Admin/GalleryAdmin.php#L69 
https://github.com/sonata-project/SonataNewsBundle/blob/75b0c68e9bb77d75f0605520eb1af12c63071b36/src/Admin/PostAdmin.php#L122
> 
> There are some more, but I'm lazy to link every code fragment ;)
> 
> > A media has a context property, not a ContextInterface one.
> 
> That's right, but the context property of a media instance is linked to the context id of the `ContextInterface`:
> 
> https://github.com/sonata-project/SonataMediaBundle/blob/961e51a0da4b0be82ca0ada2644422f918c4e6f2/src/Command/FixMediaContextCommand.php#L70-L72

^ Just for the reference ^


For me, context**ID** is misleading, because i would expect some integer or UUID value for an ID. I can change this if I'm the only one with this concerns.

wbloszyk: `Context` from MediaBundle is array:
```
sonata_media:     
    contexts:
        news:
            download:
                strategy: sonata.media.security.superadmin_strategy
                mode: http
            providers:
                - sonata.media.provider.image
                - sonata.media.provider.file
            formats:
                small: { width: 150 , quality: 95}
``` 

`Context` from ClassificationBundle is `ContextInterface`. 

IMO In both cases `news` should be identifier, not name like here: https://github.com/sonata-project/SonataMediaBundle/blob/26991351cac2b1399890685b43c4d20efc69da11/src/Block/GalleryListBlockService.php#L79

We should use `contextId` where `string` is using, `context` when `object|array` is using.

VincentLanglet: > For me, contextID is misleading, because i would expect some integer or UUID value for an ID

Id is to me an identifier

There is a `getId` method, it's not called `getContext()`.
https://github.com/sonata-project/SonataClassificationBundle/blob/cf0d56c34644d445178a54d32f77f036f2113946/src/Model/ContextInterface.php#L48

If I have a `$foo` variable, instance of ContextInterface, when I call `getId()` on it, I will call this `$contextId`, not `$context`.

So `$context` seems more misleading.

VincentLanglet (approved):

src/Entity/TagManager.php
```diff
@@ -36,12 +36,12 @@ public function getBySlug(string $slug, $context = null, ?bool $enabled = true):
...
     }
 
-    public function getByContext($context, ?bool $enabled = true): array
+    public function getByContext(string $contextId, ?bool $enabled = true): array
     {
         $queryBuilder = $this->getRepository()
             ->createQueryBuilder('t')
             ->select('t')
-            ->andWhere('t.context = :context')->setParameter('context', $context);
+            ->andWhere('t.context = :context')->setParameter('context', $contextId);
```
wbloszyk, line 44: It is pedantic but will be awesome to change it too.
```suggestion
            ->andWhere('t.context = :contextId')->setParameter('contextId', $contextId);
```

wbloszyk (approved):

PR #757, opened 2021-10-28
Remove call to unknown method

core23:
## Subject

We don't need this call, as the admin is set via https://github.com/sonata-project/SonataAdminBundle/blob/ab9f4e0771e303628d0ea3fe99b0b56cb7b07895/src/Admin/AbstractAdmin.php#L1824

## Changelog

```markdown
### Fixed
- Removed call to unknown `FormMapper::getAdmin` method 
```

### DIFF
--- a/src/Block/Service/AbstractClassificationBlockService.php
+++ b/src/Block/Service/AbstractClassificationBlockService.php
@@ -14,9 +14,9 @@ declare(strict_types=1);
 namespace Sonata\ClassificationBundle\Block\Service;
 
 use Sonata\AdminBundle\Admin\AdminInterface;
-use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Form\Type\ModelListType;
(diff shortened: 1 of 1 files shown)


phansys (dismissed):

src/Block/Service/AbstractClassificationBlockService.php
```diff
@@ -57,7 +57,6 @@ final protected function getFormAdminType(FormMapper $formMapper, AdminInterface
 
         $fieldDescription = $admin->createFieldDescription($field, $adminOptions);
         $fieldDescription->setAssociationAdmin($admin);
-        $fieldDescription->setAdmin($formMapper->getAdmin());
```
VincentLanglet, line 60: Can the FormMapper namespace be changed to the BlockBundle one ?

VincentLanglet (changes requested):

VincentLanglet (approved):

PR #759, opened 2021-11-01
Bump static analysis

VincentLanglet:
Phpstan level 8
- With version 1.0
- With strict rules

Psalm level 3


VincentLanglet:
```
------ ---------------------------------------------------------------------------------------------------------------- 
  Line   src/Document/CategoryManager.php                                                                                
 ------ ---------------------------------------------------------------------------------------------------------------- 
  226    Call to an undefined method Sonata\ClassificationBundle\Model\CategoryInterface::disableChildrenLazyLoading().  
 ------ ---------------------------------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Model/CategoryManagerInterface.php                                                                                                                                                                   
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  30     Method Sonata\ClassificationBundle\Model\CategoryManagerInterface::getRootCategoriesPager() return type with generic interface Sonata\AdminBundle\Datagrid\PagerInterface does not specify its types: T  
         💡 You can turn this off by setting checkGenericClassInNonGenericObjectType: false in your phpstan.neon.dist.                                                                                             
  37     Method Sonata\ClassificationBundle\Model\CategoryManagerInterface::getSubCategoriesPager() return type with generic interface Sonata\AdminBundle\Datagrid\PagerInterface does not specify its types: T   
         💡 You can turn this off by setting checkGenericClassInNonGenericObjectType: false in your phpstan.neon.dist.                                                                                             
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

VincentLanglet:
@core23 I bumped the static analysis

Why is there `getRootCategoriesPager` and `getSubCategoriesPager` in the interface ? We don't use them in this project. I think we should deprecate them/remove them. Do you agree ?

There is a call to `disableChildrenLazyLoading` which is not in the interface. What are the suggestion to solve this issue ? @sonata-project/contributors

core23:
> Why is there `getRootCategoriesPager` and `getSubCategoriesPager` in the interface ? We don't use them in this project. I think we should deprecate them/remove them. Do you agree ?

They exist for ~8 years (https://github.com/sonata-project/SonataClassificationBundle/commit/3f23c91612383e6d7d499b395f6f4e5320c4fdee). Maybe they were used on an old media-bundle version. 

If there is no usage, we could remove them.

> There is a call to `disableChildrenLazyLoading` which is not in the interface. What are the suggestion to solve this issue ? @sonata-project/contributors

Is there any usage for this method?

VincentLanglet:
> > There is a call to `disableChildrenLazyLoading` which is not in the interface. What are the suggestion to solve this issue ? @sonata-project/contributors
> 
> Is there any usage for this method?

It's used here: https://github.com/sonata-project/SonataClassificationBundle/blob/664ccebeeddf7e78fd606de2bce1fbd32b8c5e18/src/Entity/CategoryManager.php#L294

But I dunno why

VincentLanglet:
> > Why is there `getRootCategoriesPager` and `getSubCategoriesPager` in the interface ? We don't use them in this project. I think we should deprecate them/remove them. Do you agree ?
> 
> They exist for ~8 years ([3f23c91](https://github.com/sonata-project/SonataClassificationBundle/commit/3f23c91612383e6d7d499b395f6f4e5320c4fdee)). Maybe they were used on an old media-bundle version.
> 
> If there is no usage, we could remove them.

https://github.com/sonata-project/SonataClassificationBundle/pull/760

src/Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -31,6 +31,8 @@
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Twig\Environment;
 
+use function is_string;
```
jordisala1991: we normally not use this on Sonata packages

src/Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -155,8 +157,9 @@ public function configureSettings(OptionsResolver $resolver): void
 
     public function load(BlockInterface $block): void
     {
-        if (is_numeric($block->getSetting('categoryId'))) {
-            $block->setSetting('categoryId', $this->getCategory($block->getSetting('categoryId')));
+        $categoryId = $block->getSetting('categoryId');
+        if (is_int($categoryId) || is_string($categoryId)) {
+            $block->setSetting('categoryId', $this->getCategory($categoryId));
```
jordisala1991, line 160: you might take inspiration on how I solved for media-bundle: https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Block/FeatureMediaBlockService.php#L158-L173

VincentLanglet: There is a getCategory doing the job, I was trying to not touch the logic and just solving the phpstan/psalm issue.

src/Block/Service/AbstractCategoriesBlockService.php
VincentLanglet: solved

src/Block/Service/AbstractCategoriesBlockService.php
jordisala1991: Yep, but you could end up setting the categoryId to null, because getCategory can return null

VincentLanglet: Not when it's an int or a string

VincentLanglet: Oh I see, it can when nothing is found

VincentLanglet: @jordisala1991 I would say it's not an issue.
If the id is found, the entity is set. If the id is not found, null is set, which is correct since the value wasnt correct

jordisala1991 (dismissed):

src/Entity/BaseCategory.php
```diff
@@ -13,15 +13,8 @@
 
 namespace Sonata\ClassificationBundle\Entity;
 
-use Doctrine\ORM\PersistentCollection;
 use Sonata\ClassificationBundle\Model\Category as ModelCategory;
 
 abstract class BaseCategory extends ModelCategory
 {
-    public function disableChildrenLazyLoading(): void
```
ChristianGripp, line 21: This method was not deprecated. We need a CHANGELOG entry for this

src/Block/Service/AbstractCategoriesBlockService.php
```diff
@@ -178,16 +179,16 @@ public function getMetadata(): MetadataInterface
     }
 
     /**
-     * @param CategoryInterface|int $id
-     * @param mixed                 $default
+     * @param CategoryInterface|int|string|null $id
```
ChristianGripp, line 182: We need at least a CHANGELOG entry for added type hints

core23 (changes requested):

src/Block/Service/AbstractCategoriesBlockService.php
VincentLanglet: They were already supported, the phpdoc was just wrong

src/Entity/BaseCategory.php
VincentLanglet: Done on 3.x instead https://github.com/sonata-project/SonataClassificationBundle/pull/765

core23 (approved):

PR #764, opened 2021-11-01
Add exception checks

VincentLanglet:


src/Admin/Filter/CategoryFilter.php
```diff
@@ -64,6 +64,9 @@ public function getRenderSettings(): array
         ]];
     }
 
+    /**
+     * @throws \RuntimeException
```
ChristianGripp, line 68: Not sure if we need this for `RuntimeException`. Nearly every ORM related method could throw an exception.

VincentLanglet: `getFieldName` can throw a RuntimeException.

But this draft was made to show @jordisala1991 the exception things configuration. IMHO we should not add this for now and first concentrate ourselves on
- Bumping phpstan to level 8
- Using the strict rules
- Using the 1.0 version
For every projects.

PR #770, opened 2021-11-07
Fix phpstan build

VincentLanglet:
I think finally it"s better to have protected properties in order to allow adding annotation to them.

### DIFF
--- a/src/DependencyInjection/SonataClassificationExtension.php
+++ b/src/DependencyInjection/SonataClassificationExtension.php
@@ -105,18 +105,20 @@ final class SonataClassificationExtension extends Extension
      */
     private function registerSonataDoctrineMapping(array $config): void
     {
-        foreach ($config['class'] as $type => $class) {
+        $classes = $config['class'];
+        foreach ($config['class'] as $class) {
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/SonataClassificationExtension.php
```diff
@@ -105,28 +105,30 @@ private function configureAdmin(array $config, ContainerBuilder $container): voi
      */
     private function registerSonataDoctrineMapping(array $config): void
     {
-        foreach ($config['class'] as $type => $class) {
+        $classes = $config['class'];
+        foreach ($config['class'] as $class) {
             if (!class_exists($class)) {
                 return;
```
ChristianGripp, line 117: We should throw an exception if the class does not exist

core23 (dismissed):

src/DependencyInjection/SonataClassificationExtension.php
VincentLanglet: Didn't want to change the behavior because I don't know this package but I thought about the same thing

PR #787, opened 2021-12-04
Fix builds

jordisala1991:
## Subject

## Changelog

```markdown
### Removed
- Symfony deprecations
```

### DIFF
--- a/tests/Controller/CategoryAdminControllerTest.php
+++ b/tests/Controller/CategoryAdminControllerTest.php
@@ -261,7 +261,7 @@ final class CategoryAdminControllerTest extends TestCase
 
         if (false !== $context) {
             $contextMock = $this->getContextMock($context);
-            $this->request->query->set('context', $contextMock->getId() ?? '');
+            $this->request->query->set('context', $context);
             $this->contextManager->expects(static::any())
                 ->method('find')
                 ->willReturn($contextMock);
--- a/tests/Controller/CategoryAdminControllerTest.php
+++ b/tests/Controller/CategoryAdminControllerTest.php
@@ -261,7 +261,7 @@ final class CategoryAdminControllerTest extends TestCase
 
         if (false !== $context) {
             $contextMock = $this->getContextMock($context);
-            $this->request->query->set('context', $contextMock->getId());
+            $this->request->query->set('context', $contextMock->getId() ?? '');
             $this->contextManager->expects(static::any())
                 ->method('find')
                 ->willReturn($contextMock);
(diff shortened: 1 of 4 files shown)


tests/Controller/CategoryAdminControllerTest.php
```diff
@@ -261,7 +261,7 @@ public function testTreeAction($context, array $categories): void
 
         if (false !== $context) {
             $contextMock = $this->getContextMock($context);
-            $this->request->query->set('context', $contextMock->getId());
+            $this->request->query->set('context', $contextMock->getId() ?? '');
```
VincentLanglet:
```suggestion
            $this->request->query->set('context', $context);
```
The id is the argument of `getContextMock`

VincentLanglet (changes requested):

tests/Controller/CategoryAdminControllerTest.php
jordisala1991: Way better!

VincentLanglet (approved):

core23 (approved):

VincentLanglet:
I just tried to do the same in SonataMedia and discovered some admin was using POST request
https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Admin/BaseMediaAdmin.php#L130

Are you sure about the `query->get` thing @core23 @jordisala1991 ?

jordisala1991:
> I just tried to do the same in SonataMedia and discovered some admin was using POST request
> 
> https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Admin/BaseMediaAdmin.php#L130
> 
> 
> 
> Are you sure about the `query->get` thing @core23 @jordisala1991 ?

Not 100% sure tbh. About that post thing, it is used to know when the form is submitted AFAIK

VincentLanglet:
> > I just tried to do the same in SonataMedia and discovered some admin was using POST request
> > https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Admin/BaseMediaAdmin.php#L130
> > Are you sure about the `query->get` thing @core23 @jordisala1991 ?
> 
> Not 100% sure tbh. About that post thing, it is used to know when the form is submitted AFAIK

I'm not familiar enough with the request, but when POST request are made, persistent parameters are in the `$request->request` instead of `$request->query` no ? Doesn't it means we're loosing all the persistent parameters when doing a post request then with this PR ?

VincentLanglet:
I don't use this bundle. @core23 I think you're using it ; I would recommend to try those changes before doing another release.

core23:
> I don't use this bundle. @core23 I think you're using it ; I would recommend to try those changes before doing another release.

👍  I'll check this before the next release.

PR #838, opened 2022-04-26
Fix the creation of a new context

jerome-fix:
## Subject

I am targeting this branch, because it's a bug introduced on this [commit](https://github.com/sonata-project/SonataClassificationBundle/commit/ae001e0928c83236343319e1610512bad73f43a2) 

A new context is created... if it already exists. It is the opposite that should be done.

 

## Changelog

```markdown
### Fixed
- Create a new context only if it does not exist
```

### DIFF
--- a/src/Admin/ContextAwareAdmin.php
+++ b/src/Admin/ContextAwareAdmin.php
@@ -61,7 +61,7 @@ abstract class ContextAwareAdmin extends AbstractAdmin
         if ('' !== $contextId) {
             $context = $this->contextManager->find($contextId);
 
-            if (null !== $context) {
+            if (null === $context) {
                 $context = $this->contextManager->create();
                 $context->setEnabled(true);
                 $context->setId($contextId);


jerome-fix:
@VincentLanglet 

Into this file the comparison with NULL is systematically done instead of testing that the variable is an instance of the expected class. Is this a general choice for Sonata? If yes I can modify my RP in this sense.

`null === $context`  vs. `!$context instanceof ContextInterface `

Jérôme.

jordisala1991:
To me in this case is easier to read with null === $context, instead of seeing the ! on the `!$context instanceof ContextInterface` expression.

VincentLanglet (approved):

VincentLanglet:
> To me in this case is easier to read with null === $context, instead of seeing the ! on the `!$context instanceof ContextInterface` expression.

I agree.

When I see `$foo instanceof Foo` it makes me believe that $foo could be either instance of Foo or instanceof Bar.
For something Foo|null, I always pref to check null instead.

jerome-fix:
@VincentLanglet @jordisala1991 

I changed the PR. No opinion on the form, I do with the conventions of the project.

Jérôme

jerome-fix:
Hi,

Do you plan to publish this correction soon?

jordisala1991 (approved):

jordisala1991:
Thanks @jerome-fix . Not sure when I can do the release, will try to do it later today.

jerome-fix:
Thanks @jordisala1991

VincentLanglet:
On it https://github.com/sonata-project/SonataClassificationBundle/pull/839

PR #842, opened 2022-05-22
DevKit updates for 4.x branch

SonataCI:


VincentLanglet:
```
[critical] Uncaught PHP Exception Twig\Error\SyntaxError: "Unknown "sonata_form_canonicalize_locale_for_moment" function. Did you mean "canonicalize_locale_for_moment"?" at /home/runner/work/SonataClassificationBundle/SonataClassificationBundle/vendor/sonata-project/admin-bundle/src/Resources/views/standard_layout.html.twig line 69
```
@jordisala1991 do you have time to take a look ?

src/Admin/Filter/CategoryFilter.php
```diff
@@ -64,15 +64,6 @@ public function getRenderSettings(): array
         ]];
     }
 
-    protected function association(ProxyQueryInterface $query, FilterData $data): array
-    {
-        $alias = $query->entityJoin($this->getParentAssociationMappings());
-        $part = strrchr('.'.$this->getFieldName(), '.');
-        $fieldName = substr(false === $part ? $this->getFieldType() : $part, 1);
```
VincentLanglet, line 71: This seems wrong:
- First, `$this->getFieldType()` shouldnt be used as a field name, it's "TextType" or similar. It works because $part is never false, since there is always a `.` in `'.'.$this->getFieldName()`.
- Secondly, I don't see a reason to have a different behavior than https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Filter/Filter.php#L75-L81. If there is something to fix, it should be here.

PR #849, opened 2022-06-25
Fix deprecations for commands while maintian lazyness

jordisala1991:
## Subject

## Changelog

```markdown
### Removed
- Avoid deprecations for console commands on Symfony 6.1.
```

### DIFF
--- a/src/Command/FixContextCommand.php
+++ b/src/Command/FixContextCommand.php
@@ -18,12 +18,15 @@ use Sonata\ClassificationBundle\Model\CollectionManagerInterface;
 use Sonata\ClassificationBundle\Model\ContextInterface;
 use Sonata\ClassificationBundle\Model\ContextManagerInterface;
 use Sonata\ClassificationBundle\Model\TagManagerInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
+#[AsCommand(name: 'sonata:classification:fix-context', description: 'Generate the default context if none defined and attach the context to all elements')]
 final class FixContextCommand extends Command
 {
+    // TODO: Remove static properties when support for Symfony < 6.0 is dropped.
     protected static $defaultName = 'sonata:classification:fix-context';
     protected static $defaultDescription = 'Generate the default context if none defined and attach the context to all elements';
 
@@ -52,6 +55,8 @@ final class FixContextCommand extends Command
     public function configure(): void
     {
         \assert(null !== static::$defaultDescription);
+
+        // TODO: Remove setDescription when support for Symfony < 5.4 is dropped.
         $this->setDescription(static::$defaultDescription);
     }
 
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "php": "^7.4 || ^8.0",
         "cocur/slugify": "^4.0",
         "doctrine/collections": "^1.6",
-        "doctrine/persistence": "^2.0",
+        "doctrine/persistence": "^2.0 || ^3.0",
         "sonata-project/doctrine-extensions": "^1.13",
         "sonata-project/form-extensions": "^1.4",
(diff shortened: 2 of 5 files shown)


src/Command/FixContextCommand.php
```diff
@@ -18,12 +18,15 @@
 use Sonata\ClassificationBundle\Model\ContextInterface;
 use Sonata\ClassificationBundle\Model\ContextManagerInterface;
 use Sonata\ClassificationBundle\Model\TagManagerInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
+#[AsCommand(name: 'sonata:classification:fix-context', description: 'Generate the default context if none defined and attach the context to all elements')]
```
jordisala1991, line 26: I think this is the correct way to avoid this deprecation, wdyt @VincentLanglet ? In SonataAdminBundle we fixed it different but to me this is more correct since the attribue only gets used on Symfony 6.1, and the older versions can maintain what they have for lazy commands.

VincentLanglet: I prefer this solution over https://github.com/sonata-project/SonataBlockBundle/pull/1075 or (the SonataAdmin PR).
So the attribute solve the deprecation even if the static property is still used ?

Is there a reason to not doing this on Admin/BlockBundle or can we update the code too ?

VincentLanglet (approved):

src/Command/FixContextCommand.php
jordisala1991: We can update that code too.

src/Command/FixContextCommand.php
```diff
@@ -18,12 +18,15 @@
...
 final class FixContextCommand extends Command
 {
+    // TODO: Remove static properties when support for Symfony < 6.0 is dropped.
```
franmomu, line 29: I guess this can be removed when dropping support for Symfony < 5.4

franmomu (approved):
I'm fine with this way too, I did that way to not have to change code when dropping versions.

src/Command/FixContextCommand.php
```diff
@@ -18,12 +18,15 @@
...
     protected static $defaultName = 'sonata:classification:fix-context';
     protected static $defaultDescription = 'Generate the default context if none defined and attach the context to all elements';
```
franmomu, line 31: ~This one could be removed too~ didn't read `properties`

src/Command/FixContextCommand.php
jordisala1991: The attribute was introduced at 6.0. So it is better to remove both at 6.0.

VincentLanglet:
Can you do the PR for SonataAdmin and SonataBlock ? @jordisala1991

src/Command/FixContextCommand.php
franmomu: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Attribute/AsCommand.php Looks like it was already in 5.4

jordisala1991:
Yes, you do know if there are other bundles with commands?

VincentLanglet:
> Yes, you do know if there are other bundles with commands?

Only
https://github.com/sonata-project/SonataSeoBundle/tree/3.x/src/Command
and pagebundle. (but doesn't support SF5 yet)

src/Command/FixContextCommand.php
jordisala1991: Yes, and maybe is not correct documented, but the change where they make the AsCommand load lazy is on 6.0

jordisala1991: https://symfony.com/doc/5.4/console/commands_as_services.html#lazy-loading

vs

https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading

franmomu: I thought it was because I read it in https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description

jordisala1991: Seems like you are right even though it is not documented. Will try to update those indications on next pR.

PR #889, opened 2023-03-04
Drop older doctrine persistence, allow collections 2.0, update psalm, incresing it to level 2 and avoid deprecations

jordisala1991:
## Subject

I am targeting this branch, because this should be BC and probably fix at least one bug.

## Changelog

```markdown
### Added
- Added support for `doctrine/collections` ^2.0.

### Removed
- Removed support for `doctrine/persistence` ^2.0.

### Fixed
- Avoid deprecations for SonataAdminBundle filters.
```


src/Form/Type/CategorySelectorType.php
```diff
@@ -47,29 +48,39 @@ public function configureOptions(OptionsResolver $resolver): void
 
     /**
      * @return array<array-key, string>
+     *
+     * @phpstan-param Options<array{
```
jordisala1991, line 52: Do you know if this can be done better @VincentLanglet ?

PHPStan understand the options param has 2 keys, but it is unable to know that 'context' will be ContextInterface or null. It is mixing keys and values.

jordisala1991: I think I saw it, the issue opened by you: https://github.com/phpstan/phpstan-symfony/issues/322

src/Entity/CategoryManager.php
```diff
@@ -146,7 +146,7 @@ public function getBySlug(string $slug, ?string $contextId = null, ?bool $enable
             ->andWhere('c.slug = :slug')->setParameter('slug', $slug);
 
         if (null !== $contextId) {
-            $queryBuilder->andWhere('c.context = :context')->setParameter('context', $contextId, Types::OBJECT);
+            $queryBuilder->andWhere('c.context = :context')->setParameter('context', $contextId, Types::JSON);
```
jordisala1991: Types::OBJECT is deprecated for Types::JSON

src/Form/Type/CategorySelectorType.php
jordisala1991: We might supress the issues raised, and wait for a fix on phpstan side.

src/Document/CategoryManager.php
```diff
@@ -206,11 +206,6 @@ protected function loadCategories(ContextInterface $context): void
             $categoryId = $category->getId();
             \assert(null !== $categoryId);
             $this->categories[$contextId][$categoryId] = $category;
-
-            $parent = $category->getParent();
-            if (null !== $parent) {
-                $parent->addChild($category);
```
jordisala1991, line 212: This should have been removed when we deprecated/removed the disableLazyChildren collection here: https://github.com/sonata-project/SonataClassificationBundle/pull/765

src/Entity/CategoryManager.php
```diff
@@ -146,7 +145,7 @@ public function getBySlug(string $slug, ?string $contextId = null, ?bool $enable
             ->andWhere('c.slug = :slug')->setParameter('slug', $slug);
 
         if (null !== $contextId) {
-            $queryBuilder->andWhere('c.context = :context')->setParameter('context', $contextId, Types::OBJECT);
+            $queryBuilder->andWhere('c.context = :context')->setParameter('context', $contextId);
```
jordisala1991, line 148: This was failing when trying to query with a non mocked CategoryManager on CategoryManagerTest, wdyt? ContextId is not an object, but a string.

src/Entity/CategoryManager.php
```diff
@@ -196,11 +195,6 @@ protected function loadCategories(ContextInterface $context): void
             $categoryId = $category->getId();
             \assert(null !== $categoryId);
             $this->categories[$contextId][$categoryId] = $category;
-
-            $parent = $category->getParent();
-            if (null !== $parent) {
-                $parent->addChild($category);
```
jordisala1991, line 202: same as the other lazy comment.

src/Resources/config/doctrine/Category.mongodb.xml.skeleton
```diff
@@ -1,10 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
jordisala1991, line 1: Those skeletons were used by the EasyExtends bundle, should have been removed by 4.0

src/Block/Service/AbstractClassificationBlockService.php
```diff
@@ -66,7 +71,7 @@ final protected function getFormAdminType(FormMapper $formMapper, AdminInterface
             'required' => false,
         ], $fieldOptions);
 
-        return $formMapper->create($formField, ModelListType::class, $fieldOptions);
+        return $formMapper->getFormBuilder()->create($formField, ModelListType::class, $fieldOptions);
```
jordisala1991, line 74: Here I am trying to avoid the deprecated call to ->create(). Following same strategy as we did for blocks on PageBundle: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Block/SharedBlockBlockService.php#L148

There is not an easy way to test if the whole block works or not, but I think I will leave it for a follow up PR. My guess is that those blocks does not work, with or without this PR.

VincentLanglet (approved):

